PR #43, opened 2012-03-31
As promised, here are unit tests for lxml.html.html5parser

dairiki:
There are three commits in this pull request.   The first two should be fairly non-controversial.  The third may warrant some discussion.

The first commit fixes an unrelated test which was failing for me.

The second commit includes unit tests for html5parser.py.

The third commit addresses what I think is a bug.   When the code in html5parser generates dummy wrapper elements (e.g. a ``<div>`` to wrap multiple fragments) it currently generates non-namespaced elements.   By default the html5lib parser generates namespaced elements, so this is a bit of a mismatch.   This patch namespaces these generated wrapper elements when the parser is configured to namespace its output.

Other Notes
-----------------

I've only tested this under python 2.6.   I suspect the unit tests will need a little bodywork before they run in py3k.

I have not tested the unit tests for the XHTMLParser.   From what I can tell, html5lib.XHTMLParser is obsolete.   My version of html5lib (0.90) does not include it in any case.

### DIFF
--- a/src/lxml/html/html5parser.py
+++ b/src/lxml/html/html5parser.py
@@ -110,7 +110,7 @@ def fragment_fromstring(html, create_parent=False,
 
     if create_parent:
         if not isinstance(create_parent, _strings):
-            create_parent = 'div'
+            create_parent = _ns_prefix(parser) + 'div'
         new_root = Element(create_parent)
         if elements:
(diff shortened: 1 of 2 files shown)


scoder:
Should be superseded by recent changes in html5parser.

PR #130, opened 2014-01-25
Disable resolve_entities by default w/ one billion laughs expansion bug.

rogerhu:
Facebook recently issued its largest bug bounty to remove this security bug.

https://www.facebook.com/BugBounty/posts/778897822124446?stream_ref=10

This may result in breaking changes but should be considered an important
security enhancement.

Test fix included.

### DIFF
--- a/doc/resolvers.txt
+++ b/doc/resolvers.txt
@@ -83,7 +83,7 @@ Resolvers are registered local to a parser:
 
 .. sourcecode:: pycon
 
-  >>> parser = etree.XMLParser(load_dtd=True)
+  >>> parser = etree.XMLParser(load_dtd=True, resolve_entities=True)
   >>> parser.resolvers.add( DTDResolver() )
 
 Note that we instantiate a parser that loads the DTD.  This is not done by the
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -710,13 +710,14 @@ general document handling.
 
 .. sourcecode:: pycon
 
+    >>> parser = etree.XMLParser(resolve_entities=True)
     >>> root = etree.XML('''\
     ... <?xml version="1.0"?>
     ... <!DOCTYPE root SYSTEM "test" [ <!ENTITY tasty "parsnips"> ]>
     ... <root>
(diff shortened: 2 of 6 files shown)


plq:
Facebook's main fault here is not using Spyne. I would have protected them :D

Jokes aside, Spyne has always had numerous parser flags off due to similar security concerns. See https://github.com/arskom/spyne/blob/89b14bf00014be82933b8d7ce80af0919e6075e4/spyne/protocol/xml.py#L145 and also https://pypi.python.org/pypi/defusedxml for a good overview

https://github.com/arskom/spyne/blob/master/spyne/protocol/xml.py#L145

boxed:
Ping?

Having your package insecure by default and pointing to the documentation is not a good approach. You should point to the documentation for how to ENABLE insecure features.

sabriabri88 (approved):

sabriabri88 (approved):

sabriabri88 (approved):

sabriabri88 (approved):

sabriabri88 (approved):

sabriabri88 (approved):

clayg:
@boxed gets it - keep fighting brother!  :muscle:

scoder:
See my comments above, specifically:

> My proposal would be to add a new parser option "resolve_file_entities". If that is False, which would be the new (backwards incompatible) default, then lxml would hook into the getEntity() callback function of the SAX parser (which points to xmlSAX2GetEntity() by default AFAICT) and reject any request for parsing an external entity.

and then, after thinking this through a bit more:

> We could change the existing boolean option "resolve_entities" into one that accepts string values. A non-empty string is True, so passing something like "internal" would result in a True value in old lxml versions and be correctly interpreted in new versions. Then we could change the default from True to "internal", which would disable external file loading. Users could then fix their code by passing True instead, if they really need that feature. That would reduce the number of affected users to a minimum, while still changing the default to something safer.

scoder:
Superseded by https://github.com/lxml/lxml/pull/391

PR #145, opened 2014-11-02
Do not filter out inline data in images

jbq:
Image inline data is not necessarily Javascript so data: attribute should not be stripped out


scoder:
rejecting in favour of #162

PR #153, opened 2014-12-17
Simplify SetMixin by rebuilding it on top of MutableSet

masklinn:
SetMixin can't trivially be removed as:
* MutableSet annoyingly does not include the method form of the operators [on
  purpose](http://bugs.python.org/issue22089)
* SetMixin provides a ``copy`` which was only used internally (?) but may be
  used by third parties
Furthermore, because MutableSet and SetMixin differ in their abstract methods
(MutableSet uses discard rather than remove, and __len__ and __contains__ are
abstract) this would require significant edition of subclasses.

However rebuilding it on top of MutableSet significantly simplifies its implementation, much of which is simply aliasing the operator methods of MutableSet. This requires defining actual functions for in-place operators as the operators return `self` and the methods don't. This can also be useful to support multi-argument functions in the future if there ever is a need for it.

The diff could be simplified a bit (at the expense of final size) by leaving the old versions of `union`, `intersection`, `difference` and `symmetric_difference` as they called into the corresponding operator anyway.

`_from_iterable` was overridden to keep the current semantics of `union`/`intersection`/`difference`/`symmetric_difference` (returning a builtin `set`), MutableSet's default is to create a new object of the same type as the current one.

### DIFF
--- a/src/lxml/html/_setmixin.py
+++ b/src/lxml/html/_setmixin.py
@@ -1,4 +1,6 @@
-class SetMixin(object):
+from collections import MutableSet
+
+class SetMixin(MutableSet):
 
     """
     Mix-in for sets.  You must define __iter__, add, remove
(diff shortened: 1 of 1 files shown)


scoder:
Thanks!

PR #155, opened 2015-01-07
Fix unicode character checks.

opottone:
As per XML standard, disallow characters #xD800-#xDFFF, #xFFFE, #xFFFF
in strings.

See also
http://stackoverflow.com/questions/23184263/central-way-to-filter-invalid-unicode-chars-in-lxml


opottone:
Doesn't yet properly work on narrow builds, I'll look into it.

PR #156, opened 2015-02-14
Improve document type declaration processing

opottone:
3 bugfixes and one improvement related to document type declarations:

 * Properly escape system literal which contains a quote character "
 * Properly serialize declarations in HTML
 * In elementsoup, fix processing of everything outside the root element (document type declaration, comments, processing instructions)
 * Set document type via lxml API (without parsing a document)

### DIFF
--- a/src/lxml/html/soupparser.py
+++ b/src/lxml/html/soupparser.py
@@ -3,9 +3,11 @@ __doc__ = """External interface to the BeautifulSoup HTML parser.
 
 __all__ = ["fromstring", "parse", "convert_tree"]
 
+import re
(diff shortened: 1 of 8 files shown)


scoder:
I generally prefer telling commit descriptions rather than "go to that web page if you want to know what I did in this commit". Adding the bug number is a good thing, but only as an additional reference.

scoder:
thanks for implementing all this, it's a clear improvement.

opottone:
@scoder thanks for comments.

Sorry, I forgot to test this with python versions other than 2.7. I'll fix that and take the comments into account.

PR #157, opened 2015-02-16
Improve document type declaration processing

opottone:
4 bugfixes and one improvement related to document type declarations:

* Properly escape system literal which contains a quote character "
* Properly serialize HTML declarations
* In elementsoup, fix processing of everything outside the root element (document type declaration, comments, processing instructions)
* Set/remove document type via lxml API
* Properly serialize comments & processing instructions which precede document type declaration


scoder:
Looks good to me, thanks!

PR #159, opened 2015-02-21
Unicode & string processing

opottone:
Commits related to strings, encodings, and allowed characters.

 * Check that all strings consists of allowed XML characters (no` \ufffe`, `\uffff`, or surrogates).
 * Check that names are valid as specified in current XML standard, which is more tolerant than earlier, outdated editions (http://www.w3.org/TR/REC-xml/#NT-Name vs. http://www.w3.org/TR/2006/REC-xml-20060816/#NT-Name)
 * Python 2 expects `__repr__` to return `str`, not `unicode`, and wrong type may lead to UnicodeEncodeError. Fix this.
 * Some test contained `_str('ha\u1234\x07ho')` or similar things. The `\u` escape does not work on python 2 `''` string, unless we do some magic.

### DIFF
--- a/src/lxml/apihelpers.pxi
+++ b/src/lxml/apihelpers.pxi
@@ -1334,24 +1334,43 @@ cdef inline int isutf8(const_xmlChar* s):
         c = s[0]
     return 0
 
-cdef int check_string_utf8(bytes pystring):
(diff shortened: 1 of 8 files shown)


scoder:
Looks good to me, thanks!

PR #160, opened 2015-02-26
Failing ISOSchematron validation on report

numshub:
Standard behaviour in lxml.isoschematron.Schematron is to fail validation only on assert, ignoring any report. Failure on report can be achieved subclassing and customizing the Schematron class. Anyway an easy switch in the constructor to enable failure on report could be useful.


scoder:
I generally prefer having related changes in one commit (especially all relevant files), and also more descriptive commit comments than "update file X" or "fix bug". But anyway, this looks good enough to merge.

Thanks!

PR #161, opened 2015-02-27
isoschematron.Schematron constructor now accepts a Xpath object as validation_errors param

numshub:
Yesterday version was a bit on a rush. I also read Holger's suggestion and went back to give it a try in a proper way.

Now Schematron can be customized in two ways:
- simple customization via the validation_errors param to set a custom XPath object
- complete customization via traditional subclassing

Schematron.FAIL_ON_ASSERT_ONLY and Schematron.FAIL_ON_ASSERT_AND_REPORT are there as shortcuts to the relevant XPath objects.

This implementation goes well with the old code for both the default behavior and subclassing approach.


scoder:
looks good to me, too.

scoder:
https://github.com/lxml/lxml/commit/f0e2a42556f2c0117dd6434dddf40662681d3cf3

PR #162, opened 2015-03-04
[IMP] allow dataurls if they point to images

hbrunn:
fixes #145

While I strongly disagree to allow all dataurls, image dataurls are safe to the best of my knowledge.


scoder:
Thanks!

PR #163, opened 2015-03-05
Remove bugtrack_url because it causes warning

msabramo:
I don't think `bugtrack_url` has any effect on PyPI (you have to set the bugtrack_url in the PyPI UI I believe) and this causes a `Unknown distribution option: 'bugtrack_url'` warning to be emitted when installing or building a wheel.

    $ pip wheel lxml
    Collecting lxml
      Using cached lxml-3.4.2.tar.gz
        /Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'bugtrack_url'
          warnings.warn(msg)


scoder:
sure, why not.

PR #164, opened 2015-03-05
Friendlier error when libxml2 not installed

msabramo:
Use distutils.ccompiler.has_function to detect whether libxml2 is
installed in setupinfo.py. This allows failing earlier with a somewhat
more helpful error message if libxml2 is not installed.

On OS X, I display a tip about `xcode-select --install`. Normally I
wouldn't put platform-specific stuff in here (i.e.: I don't advise using
apt-get on Debian/Ubuntu or yum on Fedora/RedHat/CentOS) because it's a
pain to enumerate all the special cases and usually most people on Linux
platforms know their platform well enough to know how to install
packages. I made an exception for OS X and `xcode-select --install`,
because I am pretty savvy with Linux and OS X and getting things to
build, but the `xcode-select --install` thing was not obvious to me at
all. At one point, I had concluded that libxml2 must've been removed
from OS X and I would have to `brew install` it and specify environment
variables to make the compiler find it, because it's a keg-only formula,
but then I found a post on Stack Overflow, which illustrated that
`xcode-select --install` might be all it takes to get Xcode to put the
files in the proper places so that command-line tools can find them.
Since this was not obvious and a good tip, I decided to add it here.

### DIFF
--- a/setupinfo.py
+++ b/setupinfo.py
@@ -1,6 +1,6 @@
 import sys, os, os.path
 from distutils.core import Extension
-from distutils.errors import DistutilsOptionError
+from distutils.errors import CompileError, DistutilsOptionError
 from versioninfo import get_base_dir
 
 try:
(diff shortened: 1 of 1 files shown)


msabramo:
Here's the Stack Overflow post that I mentioned: http://stackoverflow.com/questions/19532125/cant-install-pil-after-mac-os-x-10-9

msabramo:
Hmmm, this doesn't seem to work right on Linux; it's not detecting when libxml2 **is** installed.

msabramo:
This works though, because `xmlXPathInit` takes no arguments:

```
>>> compiler.has_function('xmlXPathInit', include_dirs=['/usr/include/libxml2'], includes=['libxml/xpath.h'], libraries=['xml2'])
True
```

msabramo:
OK, updated with respect to your comment about checking for libxml2 only when `build_pyx.run(setup)` raises an exception.

And now the Travis CI tests are passing.

msabramo:
Any more feedback on this?

PR #165, opened 2015-04-15
Fix selector cache in ElementPath

Changaco:
Origin of the problem: 8bafbdc13ffb4fb8436eda01594780aac4735528.


scoder:
This looks embarrasingly right. Thanks!

PR #166, opened 2015-04-29
Fix undefined symbol: libiconv when using --static-deps

commiebits:
Link ordering should be specified by library dependencies (eg, libxml2
depends on libiconv) instead of os.listdir ordering


scoder:
Good catch, thanks!

PR #167, opened 2015-05-15
add keep_nsmap argument to cleanup_namespaces

tomkralidis:


tomkralidis:
Partially complete, when building XML from scratch the implementation works, when building from `etree.fromstring` the tests (locally) fail.  Any advice would be great.

tomkralidis:
@scoder: I changed `keep_nsmap` (dict) to `keep_ns_prefixes` (list).  The fix works _except_ for when the root element has no namespace.  Not sure what I'm missing.

scoder:
> The fix works _except_ for when the root element has no namespace.

That sounds like one test isn't enough here.


> Not sure what I'm missing.

I don't currently have time to look into this, so unless you can dig up
some hints here, I'd have to see when I can take a closer look.

scoder:
I think this is good enough for merging it. Any further fixes can come later.
Thanks!

scoder:
I ended up re-implementing most of it. Works now.
https://github.com/lxml/lxml/commit/9f4f87732555c13e46e48bd6f45b1db57a496b1c

PR #169, opened 2015-06-04
support for bs4.BeautifulSoup

losintikfos:
Changes to allow soapparser to attempt import from both old version of BeautifulSoup or from the new bs4.BeautifulSoup depending on current platform installation version.


scoder:
thanks. I think we should prefer bs4 over version 3 when both are available, though.

scoder:
https://github.com/lxml/lxml/commit/c1274cc5929503d48f238c8cc716b7951013db1d

PR #171, opened 2015-06-21
Input validation & encoding detection

opottone:
Here are two unrelated commits.

The first one is straightforward input validation.

The second one is about encoding detection. In progressive parsing libxml2 does not properly detect the encoding and lxml has a workaround for that. However the problem with libxml2 is that it is not used properly (can't blame you, it's not properly documented.) It's simpler if we use libxml2 as intended and drop the workaround.


opottone:
Blast, not Python 3 compatible. I'll close this and try again.

PR #173, opened 2015-08-20
fixed namespace typo

udos:
fixed a minor typo in the docs

### DIFF
--- a/doc/xpathxslt.txt
+++ b/doc/xpathxslt.txt
@@ -140,8 +140,8 @@ in the XPath expression to namespace URIs:
   ... ''')
   >>> doc = etree.parse(f)
 
-  >>> r = doc.xpath('/t:foo/b:bar',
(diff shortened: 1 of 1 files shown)


scoder:
I can't see the advantage of using 't' instead of 'a' here.

scoder:
but ok, I can live with 'a', too :)

udos:
:) good. the example in the docs required the 'a'. with the 't' it would not work ;) I should have included the link -> http://lxml.de/xpathxslt.html#namespaces-and-prefixes

with this it becomes clearer, agreed?

scoder:
Ah - now I remember why there was a 't' in there. 'a' would make it look like there was a correlation between the prefixes used in the input data and the those used in the XPath expression. It's better to keep different prefixes here, see the paragraph below the example.

udos:
uff... I understand. sorry about this!

PR #174, opened 2015-09-07
Make html5parser put elements into void namespace

sideshowbarker:
I would think that most lxml users would be happy if html5parser ignored namespaces in the same way that lxml.html does (so that they could write code with lxml that basically works the same way no matter which parser they use.

See the comment at https://bugs.launchpad.net/lxml/+bug/1191545/comments/2 from html5lib maintainer @gsnedders:

> html5lib follows what the HTML spec requires and puts HTML elements in the HTML namespace. If one wants to more-or-less match the behaviour of lxml.html, one can use the namespaceHTMLElements=False argument on html5lib.html5parser.HTMLParser and html5lib.html5parser.parse to put HTML elements in the void namespace (note that SVG and MathML elements are still put in their respective namespaces).

https://html.spec.whatwg.org/multipage/syntax.html#the-before-html-insertion-mode is where the relevant requirement is stated, in the step labeled **A start tag whose tag name is "html"**:

> Create an element for the token in the HTML namespace…

I would be extremely happy to contribute tests for this change if it is accepted.

### DIFF
--- a/doc/html5parser.txt
+++ b/doc/html5parser.txt
@@ -17,6 +17,8 @@ to the ``lxml.html`` module by providing ``fromstring()``,
 ``fragments_fromstring()`` that work like the regular html parsing
 functions.
 
+Note that, as with the ``lxml.html`` module, this module also ignores
+namespaces.
 
 Differences to regular HTML parsing
 ===================================
--- a/src/lxml/html/html5parser.py
+++ b/src/lxml/html/html5parser.py
@@ -29,7 +29,7 @@ class HTMLParser(_HTMLParser):
     """An html5lib HTML parser with lxml as tree."""
 
     def __init__(self, strict=False, **kwargs):
-        _HTMLParser.__init__(self, strict=strict, tree=TreeBuilder, **kwargs)
+        _HTMLParser.__init__(self, strict=strict, tree=TreeBuilder, namespaceHTMLElements=False, **kwargs)
 
 
 try:


scoder:
Hmm. This is a backwards incompatible change that simply changes the default for an option that users can easily pass in themselves. That comes close to deliberate code breakage.

I'd be happy to make this option more visible/accessible in the API, but I'm not sure we should change the default.

sideshowbarker:
> Hmm. This is a backwards incompatible change that simply changes the default for an option that users can easily pass in themselves. That comes close to deliberate code breakage.

I totally understand and of course it’s up to you and respect whatever decision you want to make on this.

But that said, my sense of things right now is that current situation is already sorta pretty bad, in that users are being confused by this (I know I was, and I personally know others that have been), and I feel like if it’s left as-is it’s going to just keep on confusing users for years and years and years.

So I wonder argue that it might be better to do it now, while the usage of the API is still relatively low vs what it’s eventually going to grow into.

Part of the issue here is that unlike XML authors, most HTML authors do not really know what XML namespaces are and generally never have to deal with them or worry about them, because most of the tools they use are not namespace-aware, so they never run into unexpected namespace-related issues.

I think a lot of those users are already using `lxml.html` and so are accustomed to never even needing to be aware of namespaces at all. But then at some point they hear about html5parser and hear that it’s based on html5lib and that it’s a parser that conforms to rules in the HTML spec and the way that Web browsers parse HTML, so then they decide, hey, I’m gonna give that a try.

But then the first time they try it, they run into unexpected problems because of the namespace issue, or they get serialized output filled with tons of `html:a` and `html:i`, etc., and they just stop there and say WTF and give up and never bother to try it again, and go back to using `lxml.html`.

So I would argue that the current state of things might actually be preventing many users from ever doing anything useful at all with html5parser to begin with. Which kind of saddens me, because I like having it a lot and would like for more people to be using it.

On top of all that, I would also wonder how much breakage the change would cause for users in practice.

What I mean is, I wonder what evidence we might be able to collect that there are a significant number of users whose code actually *depend* on the parser returning a namespaced HTML document vs the number of users who are already working around it by using methods that ignore the namespace; for example, by doing XPath evaluations like `//*[local-name() = "foo"]` rather than `//foo`.

> I'd be happy to make this option more visible/accessible in the API

OK, even just having that would be super great. I think it’s not at all very obvious to users right now how to pass the option.

So do you mean changes to just the docs? Or some other way from the code itself to make users more aware how to pass in the option for non-namespaced parsing?

scoder:
I think adding some visible hint to docs and docstrings would help already. Something like "while the default is to return a tree with fully qualified XHTML tags here, it if often useful in practice to use bare HTML tags instead. This can be done by setting this option."

scoder:
Sorry, wrong button. Didn't intend to close this.

PR #177, opened 2015-10-20
Clarify cleaner doc on setting `comments=False`

JocelynDelalande:
Maybe the actual way to go is to remove [this rule](https://github.com/lxml/lxml/blob/11e7f9a7d98e96ad31a1d5991b57d47472c1d143/src/lxml/html/clean.py#L308) you seem not to be sure about. The sure thing is that the effect is quite unexpected for the library user: "I cannot disable comments removing".

### DIFF
--- a/src/lxml/html/clean.py
+++ b/src/lxml/html/clean.py
@@ -109,7 +109,8 @@ class Cleaner(object):
         as they could contain Javascript.
 
     ``comments``:
-        Removes any comments.
+        Removes any comments. If you set it to False, keep in mind that 
+        ``processing_instructions`` also removes comments).
 
     ``style``:
         Removes any style tags or attributes.


scoder:
I agree that it's unexpected. Not sure if we should break code by fixing this (it's been like that forever), but it feels like we should fix it in the code, rather than documenting weird behaviour.

mlissner:
It's worth reviewing this too: https://bugs.launchpad.net/lxml/+bug/1882606

> Cleaner.clean_html() uses the default parser internally, which strips comments and PIs – before even passing the result through the cleaner. I admit that that may appear counter-intuitive. PR welcome to configure the parser based on the relevant Cleaner options.

scoder:
Closing since the Cleaner is no longer part of lxml.

PR #179, opened 2015-10-22
Treat None parameter as no parameter when creating elements

quasarj:
The E-factory won't accept None as a valid argument, but I feel like it should treat this the same as when it does not receive any arguments.

I don't know if there is a good reason it raises an exception in this case, but this small commit "fixes" it.

### DIFF
--- a/src/lxml/builder.py
+++ b/src/lxml/builder.py
@@ -218,6 +218,8 @@ class ElementMaker(object):
             get(dict)(elem, attrib)
 
         for item in children:
+            if item is None:
+                continue
             if callable(item):
                 item = item()
             t = get(type(item))


scoder:
I understand your use case and I think it's a valid one. OTOH, silently ignoring None values is not so unlikely to hide bugs on user side, e.g. forgetting to return something from a factory function or accidentally having a None value in an input list. This might just subtly change the output in ways not tested for (you wouldn't usually test for every little tag and its exact number of occurrences, would you?).

It's difficult to say for me which way is more helpful, but I'm leaning slightly towards keeping it an error. Note that we can already support your use case through the `typemap`. If you put `NoneType` into the typemap and map it to a function returning `None`, it will simply be ignored. This might be worth pointing out in the docs.

quasarj:
Good points. I was unaware of the option of using the typemap. I will check that out.

scoder:
Closing for the reasons I gave above.

PR #180, opened 2015-11-19
Add inline_style option

chripede:
Currently the style option will remove both style tags and attributes. 
This change makes each an option, just like scripts and javascript.

### DIFF
--- a/src/lxml/html/clean.py
+++ b/src/lxml/html/clean.py
@@ -112,7 +112,10 @@ class Cleaner(object):
         Removes any comments.
 
     ``style``:
-        Removes any style tags or attributes.
+        Removes any style tags.
+
+    ``inline_style``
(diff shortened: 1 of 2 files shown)


scoder:
This breaks both the API and a test. And it adds a new option without testing it. No way I can accept that.

chripede:
Was in a hurry and forgot the tests. I fixed the broken one and added a new one.

scoder:
It still breaks the API, though. If you want a new option for such a special case, it shouldn't break all code out there. It should just be a new option.

chripede:
So in order to not break the API, should I move the inline_style into the style and default inline_style to True?

scoder:
I think they should be separate options where the new one falls back to the old one if not provided. d3b73e561628998a8738b23268f4f8bf5ca98706

PR #182, opened 2015-12-12
Add tests and slightly deeper rnc2rng support

djc:


djc:
Hmm, Travis looks like the `skipif()` thing isn't quite working.

scoder:
LGTM, thanks!

PR #185, opened 2016-01-26
Fixed call to encode

peterdemin:
Otherwise I was getting following error during import:

```
File "src/lxml/lxml.etree.pyx", line 169, in init lxml.etree (src/lxml/lxml.etree.c:229940)
TypeError: encode() argument 1 must be string without null bytes, not unicode
```

Also vim auto-fixed trailing spaces.

### DIFF
--- a/src/lxml/lxml.etree.pyx
+++ b/src/lxml/lxml.etree.pyx
@@ -166,7 +166,7 @@ _initThreadLogging()
 xmlparser.xmlInitParser()
 
 # filename encoding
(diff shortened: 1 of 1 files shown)


scoder:
Hmm, I wonder why this poses a problem for you. This code is executed for everyone right on import...

Anyway, it's actually the right thing to do here, so, thanks!

peterdemin:
I did some experiments, and found out, that this one is very hard to reproduce. Here are my steps:
1) Build uwsgi wheel using Python 2.7.6
2) Install it in Python 2.7.10
3) Load application, that imports lxml.

Actually I don't understand, why it happens...

PR #186, opened 2016-02-22
Fix unichr bug on Python3

duncanmorris:
The unescape function in `soupparser.py` uses `unichr` which doesn't exist in Python 3.


scoder:
Thanks. I can't believe no-one notices these things for so long...

PR #191, opened 2016-04-08
Fix losing float precision for lxml.objectify FloatElement.

hjoukl:
Use repr() instead of str()/unicode() as 'stringification' function in the
FloatElement type registration.

This fixes LP:#1567505.

One of the new test methods doesn't fail when running the updated test suite without the fix (test_type_float_instantiation_precision) applied. This is intentional, creating a FloatElement from a string never lost precision. The test is in for completeness' sake.

### DIFF
--- a/doc/objectify.txt
+++ b/doc/objectify.txt
@@ -308,7 +308,7 @@ To simplify the generation of trees even further, you can use the E-factory:
     >>> E = objectify.E
     >>> root = E.root(
     ...   E.a(5),
-    ...   E.b(6.1),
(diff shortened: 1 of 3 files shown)


scoder:
One of the doctests is now failing in Py2.6, which doesn't have the improved `repr()` implementation for floats. I'd rather change the test than the implementation, though.

hjoukl:
Done. Should've probably enabled travis for my fork...

scoder:
Thanks!

PR #194, opened 2016-06-16
tutorial: Fix example to work with Python3

nevimov:
The builtin "basestring" type was removed in Python 3. As a consequence, this example’s code raises a NameError exception:

```
>>> for element in root.iter():
...     if isinstance(element.tag, basestring):
...         print("%s - %s" % (element.tag, element.text))
...     else:
...         print("SPECIAL: %s - %s" % (element, element.text))
```

This commit makes it work in both versions, by replacing the use of isinstance() with hasattr().

### DIFF
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -580,7 +580,7 @@ make sure only Element objects are returned, you can pass the
     >>> root.append(etree.Comment("some comment"))
 
     >>> for element in root.iter():
-    ...     if isinstance(element.tag, basestring):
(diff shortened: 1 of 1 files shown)


nevimov:
Now that I've thought about it some more, this change seems to me as a bad idea.
It makes the example's code less obvious. Perhaps, a better solution would be to simply add a relevant comment to the old code.

scoder:
I agree. A comment would do it, I guess. 82e8db8a0796f85296739e6cb23c85e1f0aa1d4b

PR #195, opened 2016-06-22
Import unichr from six.

zauddelig:
It seems cleaner to use six library.


scoder:
But why add a dependency if the code works just fine without it?

zauddelig:
you're right sir

PR #197, opened 2016-06-29
Ignore disabled form inputs

creitiv:
Disabled input elements are not in included in the emerging request in general according to [1](https://www.w3.org/TR/html5/forms.html#enabling-and-disabling-form-controls:-the-disabled-attribute) and [2](http://www.w3schools.com/tags/att_input_disabled.asp) which doesn't seems to be recognized by lxml yet.

These changes do not really implements the standard perfectly which includes also rules for which the disabled attribute would be invalid, but it's actually the way most browsers are handling it by disabling an input just when the disabled attribute is set at all independently of the actual value.

### DIFF
--- a/src/lxml/html/__init__.py
+++ b/src/lxml/html/__init__.py
@@ -1010,7 +1010,7 @@ class FormElement(HtmlElement):
         results = []
         for el in self.inputs:
             name = el.name
-            if not name:
+            if not name or 'disabled' in el.attrib:
                 continue
             tag = _nons(el.tag)
             if tag == 'textarea':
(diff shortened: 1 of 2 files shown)


scoder:
Thanks. Please provide a test.

creitiv:
Done.

It seems like for XHTML the corresponding attribute would have to be exactly `disabled="disabled"` as XHTML does not support minimized attributes. I'm not sure if lxml does really distinguish here, but I noticed that it maybe could make sense to recognize this in the `xhtml_to_html()` function in `lxml.html` by unminimizing HTML attributes when converting it to XHTML.

scoder:
I think your idea of converting these back and forth in `xhtml_to_html()` and `html_to_xhtml()` is a good one. Could you code up a pull request for that, too?

PR #198, opened 2016-07-18
Fix setting the base url for etree.Resolver.resolve_string

mvantellingen:
See https://github.com/GNOME/libxml2/blob/master/parserInternals.c#L1549
We seem to get away with only setting the _filename so that relative
url's are resolved based on the value.

Fixes https://bugs.launchpad.net/lxml/+bug/1568167

### DIFF
--- a/src/lxml/includes/xmlparser.pxd
+++ b/src/lxml/includes/xmlparser.pxd
@@ -56,6 +56,7 @@ cdef extern from "libxml/tree.h":
         const_xmlChar* base
         const_xmlChar* cur
         const_xmlChar* end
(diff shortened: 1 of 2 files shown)


scoder:
Thanks!

scoder:
A test would be nice, BTW.

scoder:
The file path might need encoding in some cases. I pushed a fix here: 5da090bdd9811ee2d8e76bb536fd84dc9d556033

mvantellingen:
Thanks, i'll add writing a unittest for this on my todo list :-)

PR #199, opened 2016-07-20
Exclude `file` field `value` from `FormElement.form_values`.

tdivis:
Similar to `submit`, `image` and `reset`, browsers don't send `file` field values in the POST when form is submitted. `FormElement.form_values` method already correctly excluded `submit`, `image` and `reset` fields, now it also excludes the `file` fields.


scoder:
Ok, seems reasonable. Thanks.

PR #201, opened 2016-07-29
allow None as an attribute value, for HTML boolean attributes

dholth:
I was writing a script to convert some tags in Angular templates to a new API, and noticed that curiously lxml could parse and re-serialize boolean tags that have no value, but could not create new ones. It turns out libxml2 supports this by accepting NULL as the value for xmlSetNsProp(), creating an attribute node with NULL children instead of a text node.

This change allows None as a value for Element.set() so that those who value attributes with no value can create the same.

Note that lxml.etree.tostring() serializes these as `<form novalidate=''></form>` while lxml.html.tostring() produces the value in the test.

### DIFF
--- a/src/lxml/apihelpers.pxi
+++ b/src/lxml/apihelpers.pxi
@@ -564,11 +564,14 @@ cdef int _setAttributeValue(_Element element, key, value) except -1:
     if not element._doc._parser._for_html:
         _attributeValidOrRaise(tag)
     c_tag = _xcstr(tag)
-    if isinstance(value, QName):
-        value = _resolveQNameText(element, value)
+    if value is None:
+        c_value = NULL
     else:
-        value = _utf8(value)
-    c_value = _xcstr(value)
+        if isinstance(value, QName):
+            value = _resolveQNameText(element, value)
+        else:
+            value = _utf8(value)
+        c_value = _xcstr(value)
     if ns is None:
         c_ns = NULL
     else:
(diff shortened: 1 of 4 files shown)


src/lxml/apihelpers.pxi
```diff
@@ -564,11 +564,14 @@ cdef int _setAttributeValue(_Element element, key, value) except -1:
     if not element._doc._parser._for_html:
         _attributeValidOrRaise(tag)
     c_tag = _xcstr(tag)
-    if isinstance(value, QName):
-        value = _resolveQNameText(element, value)
+    if value is None:
```
scoder: Shouldn't it reject `None` for XML trees and only allow it for HTML trees? The latter is the only use case I can see. XML always requires at least an empty string as value.

src/lxml/tests/test_htmlparser.py
```diff
@@ -592,6 +592,13 @@ def test_ietf_decl(self):
         self.assertEqual(self.etree.tostring(doc),
                          _bytes('<!DOCTYPE html PUBLIC "-//IETF//DTD HTML//EN">\n<html/>'))
 
+    def test_boolean_attribute(self):
+        # ability to serialize boolean attribute by setting value to None
+        form = html.Element('form')
+        form.set('novalidate', None)
+        self.assertEqual(html.tostring(form), 
+                _bytes('<form novalidate></form>'))
+
```
scoder: I think there should also be a test for the HTML parser behaviour here.

scoder:
Looks like a nice feature. A comment in the (lxml.html?) docs would be perfect.

src/lxml/apihelpers.pxi
dholth: XML serialization produces the empty string as value, even if it is None. Is there a way to tell which tree you are in?

src/lxml/tests/test_htmlparser.py
dholth: Maybe, although I didn't change it at all.

scoder: I know, but you are now making this an official feature. We should better make sure that it actually works, and keeps working.

scoder:
Thanks!

PR #202, opened 2016-08-17
manylinux: Add build-wheels.sh and wheel_manylinux target

joar:
The `wheel_manylinux` Makefile target starts [pypa's manylinux_x86_64 image](https://github.com/pypa/manylinux) and runs `tools/manylinux/build-wheels.sh` inside it.

`tools/manylinux/build-wheels.sh` iterates over the manylinux system's python installations and builds a wheel for each of them, the wheels are then put in `/io/wheelhouse`, where `/io` is the mounted git root of lxml.

- The wheel builds are static for now.
-  No wheels are built for Python 2.6 or 32-bit x86.


scoder:
Thanks!

scoder:
Wheels uploaded for lxml 3.6.1. Please give them a try.

PR #203, opened 2016-08-19
Add option to prevent default doctypes

dufferzafar:
This adds a new option to the `HTMLParser` which passes out corresponding flag to libxml.

There's another option `HTML_PARSE_NOIMPLIED` which is defined by libxml but we don't take it in as a parameter.

---

Reason why this was added:
https://github.com/mitmproxy/mitmproxy/issues/845

Post on mailing list:
https://mailman-mail5.webfaction.com/pipermail/lxml/2016-August/007738.html

### DIFF
--- a/src/lxml/parser.pxi
+++ b/src/lxml/parser.pxi
@@ -1614,6 +1614,7 @@ cdef class HTMLParser(_FeedParser):
     - remove_pis         - discard processing instructions
     - strip_cdata        - replace CDATA sections by normal text content (default: True)
     - compact            - save memory for short text content (default: True)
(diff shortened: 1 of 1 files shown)


scoder:
Thanks

scoder:
Note, BTW, that I'm free to remove this option or render it disfunctional at any time, given that I won't be breaking any tests while doing that. ;-)

dufferzafar:
@scoder I'm new to cython etc. and wanted to get some preliminary feedback before I delved into tests. I didn't really expect that it'll be merged so easily :P

Do you want me to write some tests & create a new PR?

scoder:
Please do.

PR #204, opened 2016-08-19
link with "-lrt" error on darwin (OS X)

nightsailer:
add -lt link will failed on Darwin (OS X 10.11)


gpaterno:
+1, I came up with the same patch, but useless having duplicate :)
👍

scoder:
Thanks!

PR #206, opened 2016-09-15
Fix xslt error parsing so filename and line numbers show.

lambdafu:
This is worth a :facepalm: ...

Before:

```
<string>:0:0:ERROR:XSLT:ERR_OK: compilation error, element 'foo'
<string>:0:0:ERROR:XSLT:ERR_OK: xsltStylePreCompute: unknown xsl:foo
<string>:0:0:ERROR:XSLT:ERR_OK: compilation error, element 'foo'
<string>:0:0:ERROR:XSLT:ERR_OK: xsltParseStylesheetTop: unknown foo element
```

After:

``` nohighlight
errtest.xml:3:0:ERROR:XSLT:ERR_OK: compilation error, element 'foo'
<string>:0:0:ERROR:XSLT:ERR_OK: xsltStylePreCompute: unknown xsl:foo
errtest.xml:3:0:ERROR:XSLT:ERR_OK: compilation error, element 'foo'
<string>:0:0:ERROR:XSLT:ERR_OK: xsltParseStylesheetTop: unknown foo element
```

We can't do better, as the second line in the two errors embellishes the first line and does not duplicate the file/line info.

Example:

``` python
from io import BytesIO
from lxml import etree
style = etree.parse("errtest.xml")
try:
    etree.XSLT(style)
except etree.XSLTParseError as e:
    print e.error_log
```

### DIFF
--- a/src/lxml/xmlerror.pxi
+++ b/src/lxml/xmlerror.pxi
@@ -643,10 +643,10 @@ cdef void _receiveXSLTError(void* c_log_handler, char* msg, ...) nogil:
                 if c_pos == msg + 1:
                     c_text = c_str  # msg == "%s..."
                 elif c_name_pos[0] == b'e':
-                    if cstring_h.strncmp(c_name_pos, 'element %s', 10):
+                    if cstring_h.strncmp(c_name_pos, 'element %s', 10) == 0:
                         c_element = c_str
                 elif c_name_pos[0] == b'f':
-                    if cstring_h.strncmp(c_name_pos, 'file %s', 7):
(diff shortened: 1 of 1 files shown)


lambdafu:
Oh, here is the xslt:

```
<xsl:stylesheet version="1.0"
    xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
    <xsl:foo />
</xsl:stylesheet>
```

It's really the same from the test case that's already there (but which only checks for presence of ERROR:XSLT).

scoder:
Ah, beautiful. Thanks!

PR #207, opened 2016-10-03
AppVeyor: Build Python 3.5 x64 wheels for Windows

mhils:
This PR enables the Python35-x64 target so that AppVeyor automatically generates 64 bit wheels for lxml. For example, [here](https://ci.appveyor.com/project/mhils/lxml) is the latest AppVeyor build for `mhils/lxml`, and [here](https://ci.appveyor.com/project/mhils/lxml/build/1.0.27/job/mr1qr4tgv0cyoe4w/artifacts) is the corresponding x64 wheel produced by AppVeyor.

These builds are now possible because I upgraded https://github.com/mhils/libxml2-win-binaries to also produce 64 bit binaries. It would be nice to finally see Python 3.5 Windows wheels on PyPI - if there's anything else I can do to make that happen, please let me know. All that needs to be done is 

1. Enabling AppVeyor for `lxml/lxml`
2. Changing the configuration filename in AppVeyor to `.appveyor.yml` (not `appveyor.yml`)
3. Uploading the resulting artifacts to PyPI.

The Python 2.7 wheel builder fails currently, but that is caused by https://bugs.python.org/issue27973 and will hopefully be fixed with the next 2.7 patch release.

Thanks!
Max

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,7 @@ environment:
   matrix:
     - python: 27
     - python: 35
-#   - python: 35-x64
+    - python: 35-x64
 
 install:
   - "SET PATH=C:\\Python%PYTHON%;c:\\Python%PYTHON%\\scripts;%PATH%"
--- a/buildlibxml.py
+++ b/buildlibxml.py
@@ -40,12 +40,16 @@ def download_and_extract_zlatkovic_binaries(destdir):
                     assert fn.endswith('.win32.zip')
                     libs[libname] = fn
     else:
+        if sys.maxsize > 2147483647:
+            arch = "win64"
+        else:
+            arch = "win32"
         url = "https://github.com/mhils/libxml2-win-binaries/releases/download/lxml/"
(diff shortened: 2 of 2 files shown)


redapple:
hi @mhils , thanks for this work!

I started [fiddling with AppVeyor](https://github.com/redapple/lxml/pull/2) to help https://github.com/scrapy/scrapy users on Windows. (And I am struggling.)

It may be a stupid comment but would it be possible to use https://github.com/mhils/libxml2-win-binaries for all wheels, not only Python 3.5? (I can't get rid of the FTP bug https://bugs.python.org/issue27973 for some reason)

Also, maybe https://github.com/mhils/libxml2-win-binaries could be hosted under lxml org on GitHub
And maybe even it could use AppVeyor to actually build the binaries? (again this may be stupid.)

In any case, if you (or @scoder ) need some help with wheels for Windows, I can dedicate some time (although I'm not familiar at all with Windows builds, nor AppVeyor)

PR #212, opened 2016-12-02
Tutorial: Created the children list now use it.

sam-m888:
Works both ways, although I believe this was what was meant.


scoder:
Ah, no. In fact, this section is meant to show examples for processing Element objects. Creating a list of children and iterating over an Element to process its children are two different use cases.

PR #213, opened 2016-12-02
Show real output and add example etree.HTML

sam-m888:


scoder:
That's actually not the real output (we're dealing with multi-byte encoded byte strings here, not Unicode), so I'll revert that part, but I like the second part of your change.

PR #214, opened 2016-12-03
Tutorial: Import BytesIO and mark text as bytes

sam-m888:
For python 3 fixes:
* NameError: name 'BytesIO' is not defined
* TypeError: a bytes-like object is required, not 'str'


scoder:
Thanks

PR #216, opened 2016-12-04
expose collect_ids for HTMLParser as well

plq:
It's said to lower the memory footprint of the parser at the expense of increased CPU load. Some use cases might care more about the amount memory consumed than CPU time, so it might make sense to expose this parameter.

### DIFF
--- a/src/lxml/parser.pxi
+++ b/src/lxml/parser.pxi
@@ -1596,7 +1596,7 @@ cdef class HTMLParser(_FeedParser):
     u"""HTMLParser(self, encoding=None, remove_blank_text=False, \
                    remove_comments=False, remove_pis=False, strip_cdata=True, \
                    no_network=True, target=None, schema: XMLSchema =None, \
-                   recover=True, compact=True)
+                   recover=True, compact=True, collect_ids=True)
 
     The HTML parser.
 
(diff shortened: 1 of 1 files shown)


src/lxml/parser.pxi
```diff
@@ -1615,6 +1615,8 @@ cdef class HTMLParser(_FeedParser):
     - strip_cdata        - replace CDATA sections by normal text content (default: True)
     - compact            - save memory for short text content (default: True)
     - default_doctype    - add a default doctype even if it is not found in the HTML (default: True)
+    - collect_ids        - Set this to False to avoid ever-growing id cache in
+                           long-running processes (at a bit of a performance cost)
```
scoder: That reads a bit biased. The downside is that you'd loose ID lookups, that should be mentioned instead.

plq:
did some rewording. you can chage it to your heart's content after your merge it as well, this is a fairly simple patch.

scoder:
Hmm, did you actually test if this has an impact? The code says that it shouldn't.

plq:
Well I'm trying to troubleshoot a memory leak myself and no it doesn't seem to have an impact.

plq:
As in, the leak is still there and having this flag doesn't seem to have slowed down the rate of the leak

PR #217, opened 2016-12-04
Make XMLSyntaxError have normal SyntaxError metadata

flying-sheep:
Some tools (e.g. IPython) [show a special error message](https://github.com/ipython/ipython/blob/7bd2ff815dde32552b9c55ebbad8de83366aea46/IPython/core/ultratb.py#L704-L735) like the following if common `SyntaxError` metadata is available:

```pytb
$ PYTHONPATH=src ipython
In [1]: from lxml import etree
In [2]: etree.parse('src/lxml/tests/test_broken.xml')
  File "src/lxml/tests/test_broken.xml", line 1
    <a><b></c></b></a>
             ^
XMLSyntaxError: Opening and ending tag mismatch: b line 1 and c, line 1, column 11
```

This PR adds the necessary fields `filename`, `lineno` and `offset`, while making `position` a property in order to keep it in sync with `lineno` and `offset` if those are changed.

### DIFF
--- a/src/lxml/parser.pxi
+++ b/src/lxml/parser.pxi
@@ -15,10 +15,20 @@ class ParseError(LxmlSyntaxError):
 
     For compatibility with ElementTree 1.3 and later.
     """
-    def __init__(self, message, code, line, column):
+    def __init__(self, message, code, line, column, filename = None):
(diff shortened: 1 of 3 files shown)


scoder:
Cool, really nice feature.

PR #218, opened 2016-12-21
Fix <script> and <style> text content serialization

plq:
Behold:

```
>>> from six import BytesIO
>>> from lxml import etree
>>> f = BytesIO()
>>> with etree.htmlfile(f) as xf:
...     elt = etree.Element('script')
...     elt.text = "if (a < b);"
...     xf.write(elt)
...
>>> f.getvalue()
'<script>if (a < b);</script>'
```

... which is correct.

However:

```
>>> f = BytesIO()
>>> with etree.htmlfile(f) as xf:
...     with xf.element('script'):
...         xf.write("if (a < b);")
...
>>> f.getvalue()
'<script>if (a &lt; b);</script>'
```

... which is not only incorrect, but also quite annoying :)

This patch is an attempt at fixing it.

### DIFF
--- a/src/lxml/serializer.pxi
+++ b/src/lxml/serializer.pxi
@@ -954,7 +954,14 @@ cdef class _IncrementalFileWriter:
                     if self._status > WRITER_IN_ELEMENT or content.strip():
                         raise LxmlSyntaxError("not in an element")
                 content = _utf8(content)
-                tree.xmlOutputBufferWriteEscape(self._c_out, _xcstr(content), NULL)
+
+                ns, name, _, _ = self._element_stack[-1]
+                if c_method == OUTPUT_METHOD_HTML and \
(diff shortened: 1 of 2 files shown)


src/lxml/serializer.pxi
```diff
@@ -954,7 +954,15 @@ cdef class _IncrementalFileWriter:
                     if self._status > WRITER_IN_ELEMENT or content.strip():
                         raise LxmlSyntaxError("not in an element")
                 content = _utf8(content)
-                tree.xmlOutputBufferWriteEscape(self._c_out, _xcstr(content), NULL)
+                if c_method == OUTPUT_METHOD_XML:
```
scoder: This seems reversed. Shouldn't your special casing apply *only* to HTML, instead of everything but XML?

plq: Fixed.

PR #219, opened 2016-12-27
Fix attribute misquoting

plq:
Behold:

```
>>> from lxml import html
>>> from lxml.etree import htmlfile
>>> from lxml.html.builder import E
>>> from StringIO import StringIO
>>> out = StringIO()
>>> with htmlfile(out) as f:
... with f.element("tagname", attrib={"attr": '"misquoted"'}):
... f.write("foo")
...
>>> out.getvalue()
'<tagname attr=""misquoted"">foo</tagname>'
```

Expected output:

```
'<tagname attr="&quot;misquoted&quot;">foo</tagname>'
```

https://bugs.launchpad.net/lxml/+bug/1594155

Not fixed yet but we're closer than we were yesterday :)

### DIFF
--- a/src/lxml/includes/tree.pxd
+++ b/src/lxml/includes/tree.pxd
@@ -358,7 +358,7 @@ cdef extern from "libxml/tree.h":
                                 xmlDoc* doc, xmlNode* cur, int level,
                                 int format, const_char* encoding) nogil
     cdef void xmlBufAttrSerializeTxtContent(xmlOutputBuffer *buf, xmlDoc *doc,
-                                xmlAttr *attr, const xmlChar *string) nogil
+                                xmlAttr *attr, const_xmlChar *string) nogil
     cdef void xmlNodeSetName(xmlNode* cur, const_xmlChar* name) nogil
     cdef void xmlNodeSetContent(xmlNode* cur, const_xmlChar* content) nogil
     cdef xmlDtd* xmlCopyDtd(xmlDtd* dtd) nogil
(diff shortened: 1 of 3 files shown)


plq:
So I did some research on this. This bug exists because here:

https://github.com/lxml/lxml/blob/fb23f64abd87aa79924c9db3dab0f01f050e2cb3/src/lxml/serializer.pxi#L886

the wrong string output function is used. xmlOutputBufferWriteEscape is for writing text into tag bodies, not attribute values.

I'm inclined to say that the function that we rather need to use is xmlBufAttrSerializeTxtContent, here:

https://git.gnome.org/browse/libxml2/tree/xmlsave.c?id=229d1f93ceaa5e29a1e8455e73a7f346993d20b3#n2039

As can be seen, this is a non-trivial function so I'd be reluctant to copy it to lxml source.

However, it takes an xmlAttrPtr, which we don't seem to have at that point in lxml source code.

I'm not sure how to proceed from here. Any ideas?

plq:
We can't use `xmlEscapeEntities` from xmlsave.c because 1) it's static 2) it doesn't escape anything between 0x20 and 0x80, where '"' is 0x22.

plq:
Okay, I bit the bullet and ported the whole thing. I hope I did it the right way!

scoder:
Thanks! I know, this feels verbose and redundant, but given that this code is unlikely to ever change (well, functionality-wise, I mean), I also think that it's the best way to do this.

PR #223, opened 2017-02-10
Fix AppVeyor builds for Windows

TBBle:
This fixes two build-breaking issues and adds an optimisation:
* Corrupted .appveyor.yml was causing the wrong commands to be run, failing to run `pip install`
* Python 2.7.12+ FTP download bugfix wasn't applied to the code downloading the win32 pre-build libs
* Cython doesn't currently have wheels for Python 3.6, so was adding 30-50% of the build time generating local wheels, while downloading existing wheels saves only 8-10%.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,7 +14,7 @@ install:
 
       python -m pip install -U pip wheel setuptools
 
-      pip install -r requirements.txt
+      pip install -r requirements.txt --install-option="--no-cython-compile"
 
 build_script:
   - cmd: >-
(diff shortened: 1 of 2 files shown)


TBBle:
For the Cython wheel compilation, compare https://ci.appveyor.com/project/TBBle/lxml/build/1.0.6 (using wheels for Py2.7 and Py3.5, compiling Cython for Py3.6) with https://ci.appveyor.com/project/TBBle/lxml/build/1.0.7 (uncompiled Cython on all targets).

tl;dc: The elapsed time went from 682s to 584s.

scoder:
Thanks!

PR #224, opened 2017-02-12
Skip TempPathXmlFileTestCase for the incremental XML API

TBBle:
On "Windows NT or later", you cannot reopen a file created by
`tempfile.NamedTemporaryFile` while the file-like object is still
active.

See
https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

### DIFF
--- a/src/lxml/tests/common_imports.py
+++ b/src/lxml/tests/common_imports.py
@@ -175,12 +175,16 @@ else:
 try:
     skipIf = unittest.skipIf
 except AttributeError:
-    def skipIf(condition, why,
-               _skip=lambda test_method: None,
-               _keep=lambda test_method: test_method):
+    def skipIf(condition, why):
+        def _skip(thing):
+            import types
+            if isinstance(thing, (type, types.ClassType)):
+                return type(thing.__name__, (object,), {})
+            else:
+                return None
         if condition:
             return _skip
-        return _keep
+        return lambda thing: thing
 
 
 class HelperTestCase(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


TBBle:
This fixes 24 of the current 28 test-suite failures. The other four are libxml2 gzip issues, see mhils/libxml2-win-binaries#6

scoder:
Class decorators were only added in Py2.7. They are not available in Py2.6.

TBBle:
It appears the entire ['skip'](https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures) feature was added in Python 2.7. It's already being used on two methods in this class, `test_buffering` and `test_flush`. Unless I'm overlooking something?

scoder:
Note the import, there is a generic fallback.

TBBle:
Ah, I see. Thank you. I'll see if I can get a Python 2.6-compatible version going, but I don't have it set up here yet.

TBBle:
So it turns out class decorators were introduced in Python 2.6, we just needed to return a different thing if it's a test class being skipped as opposed to a test method.

scoder:
Thanks!

TBBle:
Well, it's definitely closer... https://ci.appveyor.com/project/TBBle/lxml/build/1.0.18 has Python 2.6 and 2.7 passing, and the next release of https://github.com/mhils/libxml2-win-binaries from @mhils will fix the rest now that mhils/libxml2-win-binaries#6 has been resolved.

PR #225, opened 2017-02-12
Enable caching of libraries in build step

mhils:
This PR makes it possible to experiment with local copies of the Windows dependencies instead of downloading them all the time. As the existence of a file does not imply that it is at the latest version, dependencies are only cached if a corresponding ".keep" file exists. For example, the local copy of `iconv-latest.win32.zip` is only taken if `iconv-latest.win32.zip.keep` exists.

This is motivated by https://github.com/mhils/libxml2-win-binaries/pull/7#issuecomment-279222114 and would allow for an easy testing of compiled binaries on appveyor.

### DIFF
--- a/buildlibxml.py
+++ b/buildlibxml.py
@@ -56,9 +56,12 @@ def download_and_extract_zlatkovic_binaries(destdir):
     for libname, libfn in libs.items():
         srcfile = urljoin(url, libfn)
         destfile = os.path.join(destdir, libfn)
-        print('Retrieving "%s" to "%s"' % (srcfile, destfile))
-        urlcleanup()  # work around FTP bug 27973 in Py2.7.12+
-        urlretrieve(srcfile, destfile)
+        if os.path.exists(destfile + ".keep"):
+            print('Using local copy of  "{}"'.format(srcfile))
+        else:
+            print('Retrieving "%s" to "%s"' % (srcfile, destfile))
+            urlcleanup()  # work around FTP bug 27973 in Py2.7.12+
+            urlretrieve(srcfile, destfile)
         d = unpack_zipfile(destfile, destdir)
         libs[libname] = d
 


TBBle:
The other thing that this allows would be speeding up builds in AppVeyor. It supports a caching system for files in the build-tree. It's currently based on [file-change or REST-API triggered invalidation](https://www.appveyor.com/docs/build-cache/) but if we can get it to work correctly that would cut down build-times even further.

The simplest way to make that work would be to version the libxml2-win-binaries builds in some way, but that requires touching the lxml source to change the libxml2 binary versions used.

The cache dependency validation is based on the CRC32 of the dependency, so fetching the id number for the release (https://api.github.com/repos/mhils/libxml2-win-binaries/releases/latest) would work. That would also work if you wanted to make the release zip files versioned upstream, because the same API gives you all the download URLs that we currently manually construct.

This would also require that different platform builds have different filenames, which happens to be the case now (as the libxml2-win-binaries names are not versioned), but as part of getting VS2008 support into libxml2-win-binaries I was expecting to insert the compiler name into either the filename generated by that build script, or the directory that lxml downloads them into.

mhils:
> The other thing that this allows would be speeding up builds in AppVeyor.

The complexity you describe after this makes me think that the current approach really is fine. 😉 
The major use case for this is being able to run the lxml tests when building libxml2-win-binaries on AppVeyor.

> but as part of getting VS2008 support into libxml2-win-binaries

FWIW, Python 2 (and <=3.4) is dead to me and I wouldn't be willing to take over maintenance responsibilities for that unless it's trivial to implement. Sorry.

TBBle:
That's quite reasonable. One of the options I considered for the Python 2.x approach would be to fork the tree and do VS2008 releases myself. That of course gives a distinct download URL, but I'd still like the files-on-disk to be differently named or pathed so that .keep does the right thing.

That said, VS2008 32-bit _was_ fairly trivial, it's a source of a lot of the cleanups I've produced recently, trying to trim to the essence of the build. I suspect the 64-bit would be trivial too, but the AppVeyor VS2008 x64 build system to be broken and I haven't had a chance to track down a proper VS2008 installation to ensure the script is correct.

You can see the not-yet-cleaned-up changes at https://github.com/mhils/libxml2-win-binaries/compare/master...TBBle:vs2008#diff-f0e02e42e53ec40f5d3b8bed031ad09e

The VS2012 builds (3.0-3.4) would be harder, as you have to use the Windows SDK to build, and that was pretty painful when I tried to make it work before. And my focus is Python 2.x (with mild interest in latest 3.x). I'm here at all because there's no lxml wheels for Python 2.x after 3.6.0, and it seems like a tractable problem.

scoder:
Ok, why not. It does sound like using versioned filenames (and finding the latest one) would solve this problem more generally, though. It would make it obvious if the latest version is there or not.

PR #226, opened 2017-02-14
Add Python 2.6 to the AppVeyor build

TBBle:
Python 2.6 can't run pip with "python -m", and since we're not doing
that on the next line anyway, clearly that's not needed since we have
the right Python\scripts folder in our path.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,6 +2,7 @@ version: 1.0.{build}
 
 environment:
   matrix:
+  - python: 26
   - python: 27
   - python: 35
   - python: 35-x64
@@ -10,7 +11,7 @@ environment:
 
 install:
     - SET PATH=C:\\Python%PYTHON%;c:\\Python%PYTHON%\\scripts;%PATH%
-    - python -m pip install -U pip wheel setuptools
+    - python -m pip.__main__ install -U pip wheel setuptools
     - pip install -r requirements.txt --install-option="--no-cython-compile"
 
 build: off


.appveyor.yml
```diff
@@ -10,7 +11,7 @@ environment:
 
 install:
     - SET PATH=C:\\Python%PYTHON%;c:\\Python%PYTHON%\\scripts;%PATH%
-    - python -m pip install -U pip wheel setuptools
+    - pip install -U pip wheel setuptools
```
mhils: `python -m pip` is required on Windows if you want to update pip itself. Otherwise, `pip.exe` would be write-protected because it's in use.

TBBle: Hmm, tricky then. https://ci.appveyor.com/project/TBBle/lxml/build/1.0.17/job/8go4rk5k1mreweyd shows that pip is up-to-date on the AppVeyor images, so I'm not hitting the issue there. I'd noticed the complaint locally, but it didn't see to fail the update as long as I updated nothing else.

I'll have a look later, there might be a bounce-script for pip that can be used with `python -m`. I've seen that in some other projects.

TBBle: Easy fix: `python -m pip.__main__`. That's what `python -m <pkg>` is shorthand for, as it happens. http://stackoverflow.com/a/6630931/166389

scoder:
Worked for lxml 3.7.3. Thanks!

PR #228, opened 2017-02-16
fix html.SelectElement stripping whitespace from <option> values

ashkulz:
This is a fix for [bug #1665241](https://bugs.launchpad.net/lxml/+bug/1665241). I've tried to add tests, please let me know if any changes need to be made.

### DIFF
--- a/src/lxml/html/__init__.py
+++ b/src/lxml/html/__init__.py
@@ -1327,9 +1327,7 @@ class SelectElement(InputMixin, HtmlElement):
             if el.get('selected') is not None:
                 value = el.get('value')
                 if value is None:
-                    value = el.text or ''
-                if value:
-                    value = value.strip()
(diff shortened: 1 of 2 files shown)


scoder:
Nice, thanks!

ashkulz:
Should I submit a PR for backporting it to the `3.7` branch? A `3.7.2` release is likely to get into the upcoming Debian release (either `stretch` or `stretch-backports`) :smile:

Also, Travis CI doesn't seem to be working although it seems to be configured -- is it disabled for some reason?

scoder:
While a bug-fix, this is also a change of long-standing behaviour that does not strictly belong into a 3.7.3.

ashkulz:
Fair enough. I've submitted a PR to scrapy to document this, as I encountered it while using it.

PR #229, opened 2017-02-26
Add check for c_src_dict in _fixThreadDictPtr

tbodt:
If a document is created outside of LXML and converted to an LXML node with elementFactory, its dictionary might be NULL, which would make xmlDictOwns return -1, which counts as true. If the nodes in that document are then moved to an LXML document, fixThreadDictNsForNode would then set the href and prefix pointers to point to the dictionary, which would cause a bad free when the document gets freed.


scoder:
Thanks!

PR #232, opened 2017-03-16
Fix LP1654544

kvdveer:


src/lxml/html/html5parser.py
```diff
@@ -25,12 +24,41 @@
...
+
+    # this function creates a wrapper around the specified object, which
+    # retries when html5lib complains about the useChardet argument
+
+    def inner(self, *args, **kwargs):
+        callee = getattr(super(type(self), self), method_name)
+        try:
+            return callee(*args, **kwargs)
+        except TypeError:
+            exception = sys.exc_info()[1]
```
scoder: this should spell "`except TypeError as exception`" (lxml is Py2.6+ only now)

src/lxml/html/html5parser.py
```diff
@@ -25,12 +24,41 @@
...
+    # detected the html argument would produce unicode objects.
+    # However, there is no reasonable way to predict if html5lib will
+    # detect the argument to be unicode (all of that code is private),
+    # so we'll have to settle for a retry.
+
+    # this function creates a wrapper around the specified object, which
+    # retries when html5lib complains about the useChardet argument
+
+    def inner(self, *args, **kwargs):
+        callee = getattr(super(type(self), self), method_name)
```
scoder: This will fail if the class gets subtyped. I think a mixin class would be better here.

kvdveer:
I have addressed the issues you pointed out; thanks for catching those.

scoder:
Thanks!

scoder:
I've thought about this again and I'll back out this change. This shouldn't be done in lxml. Instead, users should explicitly request the option if they want it, and lxml should enable it as a default for byte string input otherwise.

PR #234, opened 2017-03-16
allow hyphens in mailto: hosts

jwilk:
The suspicious regexp was found using [pydiatra](https://github.com/jwilk/pydiatra).


scoder:
Good catch. Thanks!

PR #235, opened 2017-03-17
remove obsolete compat code

jwilk:
Now that lxml supports only Python 2.6+ and 3.2+, compatibility code for earlier versions can be removed.


scoder:
Thank you for your continuous efforts to clean up my code and documentation. I appreciate it.

PR #236, opened 2017-03-19
Adds a `smart_prefix` option to XPath evaluations to overcome a counter-intuitive design flaw

funkyfuture:
> Namespaces are one honking great idea -- let's do more of those!

Using XPath to locate elements is quiet cumbersome when it comes to documents that have a default namespace:

```python
>>> root = etree.fromstring('<TEI xmlns="http://www.tei-c.org/ns/1.0" xmlns:x="http://www.w3.org/2000/svg"><text><body><x:svg /></body></text></TEI>')
>>> root.nsmap
{'x': 'http://www.w3.org/2000/svg', None: 'http://www.tei-c.org/ns/1.0'}
>>> root.xpath('./text/body')
[]
>>> root.xpath('./text/body', namespaces=root.nsmap)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    root.xpath('./text/body', namespaces=root.nsmap)
  File "src/lxml/lxml.etree.pyx", line 1584, in lxml.etree._Element.xpath (src/lxml/lxml.etree.c:59349)
    evaluator = XPathElementEvaluator(self, namespaces=namespaces,
  File "src/lxml/xpath.pxi", line 261, in lxml.etree.XPathElementEvaluator.__init__ (src/lxml/lxml.etree.c:170589)
  File "src/lxml/xpath.pxi", line 133, in lxml.etree._XPathEvaluatorBase.__init__ (src/lxml/lxml.etree.c:168702)
  File "src/lxml/xpath.pxi", line 57, in lxml.etree._XPathContext.__init__ (src/lxml/lxml.etree.c:167658)
    _BaseContext.__init__(self, namespaces, extensions, error_log, enable_regexp,
  File "src/lxml/extensions.pxi", line 84, in lxml.etree._BaseContext.__init__ (src/lxml/lxml.etree.c:156529)
    if namespaces:
TypeError: empty namespace prefix is not supported in XPath
```

This is a [well](http://lxml.de/xpathxslt.html) [documented](http://lxml.de/FAQ.html#xpath-and-document-traversal) [issue](http://git.net/ml/python-lxml-devel/2009-08/msg00039.html) ([also here](https://stackoverflow.com/questions/8053568/how-do-i-use-empty-namespaces-in-an-lxml-xpath-query)) and is [commonly solved](https://stackoverflow.com/questions/31177707/parsing-xml-containing-default-namespace-to-get-an-element-value-using-lxml) by manipulating the namespace mapping with an ad-hoc prefix - which loses the information what the default namespace was unless preserved - and adding that to XPath expressions. ([another hack](https://stackoverflow.com/questions/5572247/how-to-find-xml-elements-via-xpath-in-python-in-a-namespace-agnostic-way), [stdlib as well](http://www.goodmami.org/2015/11/04/python-xpath-and-default-namespaces.html) with some insights)

But this solution doesn't play well in generalising code like adapter classes where it becomes tedious and error prone because XPath expressions are not always identical (did i mention they are counter-intuitive to type?) and keeping track of namespace mappings across loosely coupled code elements introduces boilerplates. 

Ultimately, the interplay of document namespaces and XPath expressions is everything but pythonic and rather complicated than complex, though
 
> There should be one-- and preferably only one --obvious way to do it.

The root of this issue is caused by [a flaw](https://lists.w3.org/Archives/Public/www-dom/2002JulSep/0159.html) in the [XPath 1.0](https://www.w3.org/TR/xpath/#node-tests) specs that `libxml2` follows in its implementation:

> A QName in the node test is expanded into an expanded-name using the namespace declarations from the expression context. This is the same way expansion is done for element type names in start and end-tags except that the default namespace declared with xmlns is not used: **if the QName does not have a prefix, then the namespace URI is null** (this is the same way attribute names are expanded). It is an error if the QName has a prefix for which there is no namespace declaration in the expression context.

While [XML namespaces](https://www.w3.org/TR/REC-xml-names/#ns-decl) actually have a notion of an unaliased default namespace:

> If the attribute name matches DefaultAttName, then the namespace name in the attribute value is that of the default namespace in the scope of the element to which the declaration is attached.

[XPath 2.0](https://www.w3.org/TR/xpath20/#node-tests) did eventually fix this:

> A QName in a name test is resolved into an expanded QName using the statically known namespaces in the expression context. It is a static error [err:XPST0081] if the QName has a prefix that does not correspond to any statically known namespace. **An unprefixed QName, when used as a name test on an axis whose principal node kind is element, has the namespace URI of the default element/type namespace in the expression context**; otherwise, it has no namespace URI.

There's no XPath 2.0 implementation with Python bindings around (well, there is one to XQuilla that returns raw strings and is far off `lxml`'s capabilities), and it is very unlikely there's one to be implemented as the extension as a whole is a lot - which probably no one needs outside the XQuery/XSLT scene. [`libxml2` didn't intend to](https://www.mail-archive.com/xml@gnome.org/msg04082.html) ten years ago, but hey, [looking for a thesis](https://diplomky.redhat.com/topic/show/190/xpath-20-support-to-libxml) to write?

Thus I propose to backport that bug fix from XPath 2.0 to `lxml`'s XPath interfaces with an opt-in `smart_prefix` option without considering the whole standard as  

> practicality beats purity.

Behind the scenes the ad-hoc prefix 'solution' described above is applied, but completely hidden from the client code.

This pull request demonstrates the design and isn't completed yet, at least these issues still need to be addressed:

- documentation
- predicates are handled rather hackish and i have doubts that it works with more complex predicates
  - i'd appreciate test proposals for practical examples with such
  - support for predicates with the `smart_prefix` option could be dropped altogether, finer-grained selection is possible with Python and probably a common usage
- should this even be the default behavior with opt-out? afaict it wouldn't break any code as supplying a namespace map with a default namespace (mapped to `None`) is currently invalid 
  - i'd keep it out of XSLT anyway
- should result elements from such queries have a property that stores the option? so later calls on `.xpath()` of these elements would behave the same if no `smart_prefix` option is provided
- can `regex.h` be used directly from Cython, but that's not specific to this here

btw, this is the first time i used Cython and my C usage was long ago, i'm happy about every feedback for improvements.

Now, let's have some fun:

```python
>>> root = etree.fromstring('<TEI xmlns="http://www.tei-c.org/ns/1.0" xmlns:x="http://www.w3.org/2000/svg"><text><body><x:svg /></body></text></TEI>')
>>> root.nsmap
{None: 'http://www.tei-c.org/ns/1.0', 'x': 'http://www.w3.org/2000/svg'}
>>> root.xpath('./text/body', namespaces=root.nsmap, smart_prefix=True)
[<Element {http://www.tei-c.org/ns/1.0}body at 0x7f8e655587c8>]
>>> root.xpath('./text/body', smart_prefix=True)
[<Element {http://www.tei-c.org/ns/1.0}body at 0x7f8e655587c8>]
```

(oh, the inplace build option fails on my local machine without a helpful message. does anyone have a hint on that?)

### DIFF
--- a/src/lxml/etree.pyx
+++ b/src/lxml/etree.pyx
@@ -1566,14 +1566,15 @@ cdef public class _Element [ type LxmlElementType, object LxmlElement ]:
         return _elementpath.iterfind(self, path, namespaces)
 
     def xpath(self, _path, *, namespaces=None, extensions=None,
-              smart_strings=True, **_variables):
+              smart_strings=True, smart_prefix=False, **_variables):
         u"""xpath(self, _path, namespaces=None, extensions=None, smart_strings=True, **_variables)
 
         Evaluate an xpath expression using the element as context node.
         """
         evaluator = XPathElementEvaluator(self, namespaces=namespaces,
                                           extensions=extensions,
-                                          smart_strings=smart_strings)
+                                          smart_strings=smart_strings,
+                                          smart_prefix=smart_prefix)
         return evaluator(_path, **_variables)
 
     def cssselect(self, expr, *, translator='xml'):
@@ -2253,7 +2254,7 @@ cdef public class _ElementTree [ type LxmlElementTreeType,
         return root.iterfind(path, namespaces)
 
     def xpath(self, _path, *, namespaces=None, extensions=None,
-              smart_strings=True, **_variables):
+              smart_strings=True, smart_prefix=False, **_variables):
         u"""xpath(self, _path, namespaces=None, extensions=None, smart_strings=True, **_variables)
 
         XPath evaluate in context of document.
@@ -2274,7 +2275,8 @@ cdef public class _ElementTree [ type LxmlElementTreeType,
         self._assertHasRoot()
         evaluator = XPathDocumentEvaluator(self, namespaces=namespaces,
                                            extensions=extensions,
-                                           smart_strings=smart_strings)
+                                           smart_strings=smart_strings,
+                                           smart_prefix=smart_prefix)
         return evaluator(_path, **_variables)
 
     def xslt(self, _xslt, extensions=None, access_control=None, **_kw):
--- a/src/lxml/extensions.pxi
+++ b/src/lxml/extensions.pxi
@@ -36,6 +36,7 @@ cdef class _BaseContext:
     cdef dict _function_cache
     cdef dict _eval_context_dict
     cdef bint _build_smart_strings
+    cdef bytes _default_ns_prefix
     # for exception handling and temporary reference keeping:
     cdef _TempStore _temp_refs
     cdef set _temp_documents
@@ -46,8 +47,8 @@ cdef class _BaseContext:
         self._xpathCtxt = NULL
 
     def __init__(self, namespaces, extensions, error_log, enable_regexp,
-                 build_smart_strings):
-        cdef _ExsltRegExp _regexp 
+                 build_smart_strings, use_smart_prefix):
+        cdef _ExsltRegExp _regexp
         cdef dict new_extensions
         cdef list ns
         self._utf_refs = {}
@@ -71,7 +72,10 @@ cdef class _BaseContext:
                     new_extensions[(ns_utf, name_utf)] = function
             extensions = new_extensions or None
 
+        self._default_ns_prefix = b''
         if namespaces is not None:
+            if use_smart_prefix:
+                self._default_ns_prefix, namespaces = self._name_default_ns_prefix(namespaces)
             if isinstance(namespaces, dict):
                 namespaces = namespaces.items()
             if namespaces:
@@ -109,7 +113,8 @@ cdef class _BaseContext:
         else:
             namespaces = None
         context = self.__class__(namespaces, None, self._error_log, False,
-                                 self._build_smart_strings)
+                                 self._build_smart_strings, False)
+        context._default_ns_prefix = self._default_ns_prefix
         if self._extensions is not None:
             context._extensions = self._extensions.copy()
         return context
@@ -128,6 +133,32 @@ cdef class _BaseContext:
             # use C level refs, PyPy refs are not enough!
             python.Py_INCREF(utf)
         return utf
+
+    def _name_default_ns_prefix(self, namespaces):
+        namespaces = dict(namespaces)
+        if None not in namespaces:
+            return b'', namespaces
+
+        default_namespace = namespaces.pop(None)
+        default_prefix = self._generate_prefix(namespaces)
+        namespaces[default_prefix] = default_namespace
+        return default_prefix, namespaces
+
+    cdef bytes _generate_prefix(self, object namespaces):
+        prefixes = set(_utf8(x) for x in namespaces)
+        cdef:
+            bytes pool
+            int ctr
+            int ptr
+        pool, ctr, ptr = b'_xyzabc', 1, 0
+
+        while pool[ptr:ptr+1] in prefixes:
+            ptr += 1
+            if ptr == 7:
+                ptr = 0
+                ctr += 1
+
+        return pool[ptr:ptr+1] * ctr
 
     cdef void _set_xpath_context(self, xpath.xmlXPathContext* xpathCtxt):
         self._xpathCtxt = xpathCtxt
--- a/src/lxml/extensions.pxi
+++ b/src/lxml/extensions.pxi
@@ -243,11 +243,11 @@ cdef class _BaseContext:
                 xpath.xmlXPathRegisterNs(self._xpathCtxt,
                                          _xcstr(prefix_utf), NULL)
             del self._global_namespaces[:]
-    
+
     cdef void _unregisterNamespace(self, prefix_utf):
         xpath.xmlXPathRegisterNs(self._xpathCtxt,
                                  _xcstr(prefix_utf), NULL)
-    
+
     # extension functions
 
     cdef int _addLocalExtensionFunction(self, ns_utf, name_utf, function) except -1:
--- a/src/lxml/tests/test_xpathevaluator.py
+++ b/src/lxml/tests/test_xpathevaluator.py
@@ -202,6 +202,38 @@ class ETreeXPathTestCase(HelperTestCase):
             TypeError,
             root.xpath, '//b', namespaces={'': 'uri:a'})
 
+    def test_xpath_smart_prefix(self):
+        tree = self.parse('<TEI xmlns="http://www.tei-c.org/ns/1.0" xmlns:x="http://www.w3.org/2000/svg"><text><body><x:svg /></body></text></TEI>')
+
+        for x in (tree, tree.getroot()):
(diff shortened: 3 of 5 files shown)


flying-sheep:
> There's no XPath 2.0 implementation with Python bindings around

no, but there’s [py-dom-xpath](https://github.com/macropin/py-dom-xpath-redux):

```py
#parse the document into a DOM tree
rdf_tree = xml.dom.minidom.parse("install.rdf")
#read the default namespace and prefix from the root node
context = xpath.XPathContext(rdf_tree)

name    = context.findvalue("//em:id", rdf_tree)
version = context.findvalue("//em:version", rdf_tree)

#<Description/> inherits the default RDF namespace
resource_nodes = context.find("//Description/following-sibling::*", rdf_tree)
```

funkyfuture:
> should result elements from such queries have a property that stores the option?

after working with it for a shor while i realised that this is certainly desirable.

funkyfuture:
@scoder as i intend to release a library that depends on this feature soon, i'd appreciate some feedback what you think about this issue and the proposal.

scoder:
This PR crossed my path (pun intended) a couple of times already, but since there is a lot of text in the description, I needed several attempts at trying to understand what problem this solves. And I still don't get it.

It is generally a bad idea to use something like `root.nsmap` for an XPath expression, because it makes the code depend on an exact data representation and will simply fail if the sender decides to use a different prefix. In general, the prefixes used in an XML document should be ignored, and the code should be independent from them. Therefore, I cannot see how it is a problem that some documents use a default namespace. Just ignore that fact when writing your XPath expressions. Isn't that enough to solve the problem at hand?

funkyfuture:
> And I still don't get it.

hehe, a collegue and me didn't get your response either completely. it's propably a cultural issue as we don't have a cs background but rather are dealing with xml in the context of the humanities.

> It is generally a bad idea to use something like root.nsmap for an XPath expression

iirc the patch doesn't do this, it always relates to the `nsmap` of the element whose `xpath` method is called. in the example above this happens to be identical with the root element and thus the `nsmap` attribute is the same.

> will simply fail if the sender decides to use a different prefix

can you specify what a sender is? in our context prefixes are commonly canonical and nonetheless it would be our client code's responsibility to ensure that either the xpath expressions fit the target document's prefixes or to prepare the document accordingly. (correct me if i'm wrong, but following your argumentation as i understood, there should be no `nsmap` property at all? and `xpath` shouldn't allow the use of prefixes at all.)

yet, this patch addresses the case when a document uses multiple namespaces of which one is the 'default' / without a prefix (which is okay for the XML specs that `libxml2` implements) and that namespace is ought to be targeted with an XPath 1.0 expression. 

> Therefore, I cannot see how it is a problem that some documents use a default namespace. Just ignore that fact when writing your XPath expressions. Isn't that enough to solve the problem at hand?

no, i tried to show in the first code example that it cannot be ignored. given the document in that example, how would you query the `body` element in that case?

let me emphasize that i'm aware that one can get results with verbose expressions, but the focus is on usability here (as resolved with XPath 2.0). it's Python code that uses `lxml` after all.

one user story might also be a set of documents with varying namespaces (`https://foo.org/1.0`, `https://foo.org/1.1`, …) but stable prefixes where the XPath expressions with prefixes are more portable in various scripts.

acarapetis:
> one user story might also be a set of documents with varying namespaces (https://foo.org/1.0, https://foo.org/1.1, …) but stable prefixes where the XPath expressions with prefixes are more portable in various scripts.

This is exactly one of my stories right now - I'm dealing with varying versions of an XML schema, but always default namespaced. I currently resort to something like this, which feels slightly better than hacking out the namespace entirely (and having to then replace it manually after modifications in XML -> XML use-cases):

- Pull out the default xmlns, e.g. `ns = root.nsmap[None]`
- Every time I need an xpath query, do something like `root.xpath('//d:parent[d:child]', namespaces={'d':ns})`. The verbosity involved in passing namespaces is fairly easily mitigated with a helper function or some monkey patching, but having to put the dummy prefix in every element name in the xpath query is a real pain (and, as discussed above, requires parsing xpath to truly hide).

It really would be great if a straightforward solution was made available by lxml itself.

scoder:
>> Therefore, I cannot see how it is a problem that some documents use a default namespace. Just ignore that fact when writing your XPath expressions. Isn't that enough to solve the problem at hand?

> no, i tried to show in the first code example that it cannot be ignored. given the document in that example, how would you query the body element in that case?

```
>>> from lxml import etree
>>> root = etree.fromstring('<TEI xmlns="http://www.tei-c.org/ns/1.0" xmlns:x="http://www.w3.org/2000/svg"><text><body><x:svg /></body></text></TEI>')
>>> root.xpath('./t:text/t:body', namespaces={"t": "http://www.tei-c.org/ns/1.0"})
[<Element {http://www.tei-c.org/ns/1.0}body at 0x7fa3b8e59b90>]
```

It's important to keep your code working regardless of what namespace prefixes a document defines internally. If whoever provides the document decides to not use a default namespace, but the prefixes "honk" and "bloat", then this code will still work perfectly well. Yours will just attempt to do totally unnecessary work, and then fail.

> one user story might also be a set of documents with varying namespaces (https://foo.org/1.0, https://foo.org/1.1, …) but stable prefixes where the XPath expressions with prefixes are more portable in various scripts.

Sorry, but that example makes no sense at all. Prefixes have no meaning outside of a specific document context. They are purely for readability.

I'm closing this PR since it seems to stem from a pure misunderstanding, not an actual need. If there is anything to improve in the documentation in order to reduce the risk of future misunderstandings, I'd be happy about a PR for that.

scoder:
@acarapetis, if you want to do namespace-agnostic processing, you can either discard the namespaces from the tags after/during parsing, or use namespace wildcards, as described in the documentation: https://lxml.de/tutorial.html#namespaces

acarapetis:
@scoder: Unless I'm missing something, namespace wildcards don't work in `el.xpath()`. For simple queries where the limited syntax supported by `el.findall()` is sufficient, there's no problem, since `findall` supports empty namespace prefixes.

After some thought, I do think you're right that relying on certain a prefix (even the empty prefix) in the input document is not really coherent. I think I'm going to adjust my code to check every prefix in `root.nsmap` for a version of my schema rather than just raking `root.nsmap[None]`.

That being said, I still think it's a very common use case to be parsing a document that uses exactly one schema (though the exact schema URL may vary due to versioning), and it would be nice to have an easy way to tell lxml's xpath that all element names referenced come from this schema.

scoder:
For most needs, except for very selective path expressions, `.find*()` is nicer, easier and faster than `.xpath()` anyway. Missing expression features are, well, missing features. PRs welcome. See `_elementpath.py`.

> the exact schema URL may vary due to versioning

Versioning via namespace is ok as long as the namespace changes only for incompatible/major revisions. Changing the namespace for compatible evolutions is a bad idea. (Note: I'm aware that bad ideas are not uncommon in XML data formats.)

> it would be nice to have an easy way

I mentioned two ways. Whether they are easy or not certainly depends on the task at hand. Adding more features to `ElementPath` seems like a good way forward for now. Note that wildcards also work for tree iteration (because, why wouldn't they).

PR #238, opened 2017-04-11
properly enable in-code hex entity encoder

matejcik:
Function `_write_attr_string` improperly handles the fact that signedness of `char` is implementation- and platform-defined. Furthermore, in x86-land, the default is signed char, so `cur[0] >= 0x80` is always false and the whole branch is dead code.

That means that function `xmlSerializeHexCharRef` never actually executed and the encoding was happening magically inside libxml. (This is evidenced by the fact that `xmlSerializeHexCharRef` generates hex-based entities, but `test_attribute_quoting_unicode` was checking for decimal-based encoding. With that said, I'm very curious if commit 7eef15864c9dfdc74ac6182ce15fea07fe600af1 ever actually did anything?)

This PR forces using unsigned chars where appropriate, fixes incorrect `xmlOutputBufferWrite` call, and updates the testcase to test for hex-based encoding.

### DIFF
--- a/src/lxml/serializer.pxi
+++ b/src/lxml/serializer.pxi
@@ -481,6 +481,7 @@ cdef unsigned char *xmlSerializeHexCharRef(unsigned char *out, int val):
 cdef _write_attr_string(tree.xmlOutputBuffer* buf, const char *string):
     cdef const char *base
     cdef const char *cur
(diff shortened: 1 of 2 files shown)


scoder:
Thanks!

PR #249, opened 2017-05-20
Use ftplib when urllib fails

Bachmann1234:
Submitting this as a separate PR. If I get the separate repo for wheel building working this commit will be needed.


Bachmann1234:
Originally from https://github.com/lxml/lxml/pull/247 which will likely be closed in the near future

PR #251, opened 2017-07-28
Soup hyphen in comment

mozbugbox:
Fix a couple of issues with soupparser

### DIFF
--- a/src/lxml/html/soupparser.py
+++ b/src/lxml/html/soupparser.py
@@ -255,9 +255,7 @@ def _init_node_converters(makeelement):
 
     @converter(Comment)
     def convert_comment(bs_node, parent):
-        if "--" in bs_node or bs_node.endswith("-"):
-            bs_node = bs_node.replace("-", "&#45;")
-        res = etree.Comment(bs_node)
+        res = html.HtmlComment(bs_node)
         if parent is not None:
             parent.append(res)
         return res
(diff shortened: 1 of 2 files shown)


scoder:
Thanks. Could you come up with a test that shows why the "--" special handling is necessary? At first glance, it seems wrong to me to *add* a character escape in text content (which is not serialised XML).

mozbugbox:
Something like
```html
<html>
<body>
<div> Test </div>
<!-- This is -- a comment with double hyphen -->
</body>
</html>

```

When parsing with soupparser, exception raised 
```
 raise ValueError("Comment may not contain '--' or end with '-'")
```
The lxml code that raise the exception is in `lxml.etree.pyx`, `Comment()`.

mozbugbox:
hmm... It seems lxml.html can parse comment with 2 hyphens with no problem.
```python
from lxml.html import soupparser
from lxml import html

content = "<html><!-- something with -- 2 hyphen --></html>"

tree = html.fromstring(content)
print(html.tostring(tree))

tree = soupparser.fromstring(content)
print(html.tostring(tree))
```

scoder:
I think the soup parser is simply using the wrong way to instantiate an HTML comment:
```
>>> lxml.html.HtmlComment("something with -- 2 hyphen")
<!--something with -- 2 hyphen-->
```

mozbugbox:
aha, thanks for the advice. Pull updated

scoder:
Could you please also add tests to `test_elementsoup.py` ?

PR #252, opened 2017-07-29
Fix make test3

mozbugbox:


src/lxml/xsltext.pxi
```diff
@@ -225,13 +225,13 @@ cdef void _callExtensionElement(xslt.xsltTransformContext* c_ctxt,
             message = python.PyBytes_FromFormat(
                 "Error executing extension element '%s': %s",
                 c_inst_node.name, _cstr(e))
-            xslt.xsltTransformError(c_ctxt, NULL, c_inst_node, message)
+            xslt.xsltTransformError(c_ctxt, NULL, c_inst_node, "%s", message)
```
scoder: This should probably be smarter and pass the message unformatted to libxslt, see the corresponding 'parser' in
https://github.com/lxml/lxml/blob/e70dad61252efc8fb13e362db216fe1ac0149090/src/lxml/xmlerror.pxi#L643

scoder:
Fixes the problem at hand. Thanks!

PR #253, opened 2017-08-05
Added caveat to dict-of-dicts code

SKalt:
The code given to transform an element to a dict of dicts will overwrite information in parallel tags, so `<a><b>c</b><b>d</b></a>` will become `{'a':{'b':'d'}}`.  I've taken a stab at writing a replacement that places duplicated tags in arrays [here](https://gist.github.com/SKalt/9fdef848e4917538bd53a7d2368c1a9f); suggestions appreciated.

### DIFF
--- a/doc/FAQ.txt
+++ b/doc/FAQ.txt
@@ -365,6 +365,8 @@ I'm glad you asked.
          return element.tag, \
                 dict(map(recursive_dict, element)) or element.text
 
+Note `recursive_dict` will overwrite any data contained in siblings
+with the same tag.
 
 Why does lxml sometimes return 'str' values for text in Python 2?
 -----------------------------------------------------------------


scoder:
Thanks. Regarding a better implementation, remember that this is just an FAQ page and not a library or tool box. I don't think  there is a one-size-fits-all conversion. If the XML format is not intended to be JSON conforming by design, users would probably end up applying one format quirk fix or the other at some point. Giving them an idea and a simple example should be enough to get them going. (Worked for you, at least ;) )

scoder:
Also note that the first line of that FAQ section is supposed to indicate a bit of sarcasm. I've extended the disclaimer some more in 82675f9d03a180471056e9fd629e0f13ce4085a1.

SKalt:
Looks good, thanks for the quick response!

PR #257, opened 2017-11-26
Document that the tag argument can be a sequence of tags.

cjerdonek:
This is to better document that `iterparse()` and `iterwalk()` also accept a sequence of tags.


scoder:
Thanks.

PR #258, opened 2018-01-03
Include -isysroot linker arg on macos

frgtn:
Adds -isysroot linker flag when building extensions on darwin target.
Uses `xcrun --show-sdk-path` to determine isysroot value.

This is a minimal patch that allows building `lxml` on MacOS Sierra 10.12.x with XCode 9.

Based on suggestions found in https://forums.developer.apple.com/thread/87829
and addresses https://bugs.launchpad.net/lxml/+bug/1727864

### DIFF
--- a/setupinfo.py
+++ b/setupinfo.py
@@ -124,6 +124,9 @@ def ext_modules(static_include_dirs, static_library_dirs,
         ])
     _library_dirs = _prefer_reldirs(base_dir, library_dirs(static_library_dirs))
     _cflags = cflags(static_cflags)
+    _ldflags = []
+    if sys.platform == 'darwin':
+        _ldflags.extend(['-isysroot', get_xcode_isysroot()])
     _define_macros = define_macros()
     _libraries = libraries()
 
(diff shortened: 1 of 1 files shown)


frgtn:
As far as I know `xcrun` is only available on MacOS when XCode (along with command line utilities) is installed. XCode is needed to build C extensions so if users are able to build extensions they should already have `xcrun` installed.

Here is a man page for `xcrun`: https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/xcrun.1.html

The `xcrun` command will only be invoked in MacOS (`sys.platform == 'darwin') so it shouldn't cause problems for other platforms.

kespindler:
I just encountered this issue on my mac, and this PR is what fixed it. Specifically, I had to do:

```
pip3 install cython
pip3 install git+git://github.com/frgtn/lxml.git@5bfc9f03ffb6458ed579f4166cc9825ed7d46315
```

Really appreciate the PR!

kespindler:
👍

PR #259, opened 2018-01-26
Support using pkg-config for xml2-config/xslt-config tools

TingPing:
Those tools are not supported on some distributions where pkg-config is recommended and otherwise works fine.


TingPing:
> Do I understand it correctly that you are are passing "pkg-config" as "xslt-config" here?

Yes.

> Could you please add some documentation here that explains the why, when and how?

Will do.

TingPing:
@scoder Bump, the docs were updated.

PR #260, opened 2018-02-21
Check maxlength in value setters

cschramm:
TextareaElement and InputElement now check that a given value does not exceed the maxlength parameter if it is set.

### DIFF
--- a/src/lxml/html/__init__.py
+++ b/src/lxml/html/__init__.py
@@ -1290,6 +1290,9 @@ class TextareaElement(InputMixin, HtmlElement):
     @value.setter
     def value(self, value):
         del self[:]
+        maxlength = self.get('maxlength')
+        if maxlength and len(value) > int(maxlength):
+            raise ValueError("Value %r exceeds maxlength %r" % (value, maxlength))
         self.text = value
 
     @value.deleter
@@ -1647,6 +1650,12 @@ class InputElement(InputMixin, HtmlElement):
                 if isinstance(value, basestring):
                     self.set('value', value)
         else:
+            maxlength = self.get('maxlength')
+            if maxlength and self.type in \
+                    ['text', 'email', 'search', 'password', 'tel', 'url'] and \
+                    len(value) > int(maxlength):
+                raise ValueError("Value %r exceeds maxlength %r" % (value, maxlength))
+
             self.set('value', value)
 
     @value.deleter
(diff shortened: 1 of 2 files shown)


scoder:
Hmm, I'm actually not so sure that I want this kind of validation in the setter. It's one thing to do HTML defined structural updates like "uncheck the current radio group item when selecting a different one", but rejecting user input just because a form field says that it doesn't want it seems overly restrictive. Why should users not be allowed to send arbitrary input into a form?

If provided at all, this functionality would have to be optional, which suggests some explicit `FormElement.validate()` method that takes into account as much of the form setup as possible, e.g. the format of email input fields, min/max number values, field lengths, and what not.

And I'd be surprised if this didn't exist yet as a separate library somewhere...

cschramm:
`RadioGroup` and `SelectElement` restrict the values as well.

If you fill the form to then get all values from `form_values`, you can set arbitrary values by modifying the resulting values afterward. If you're using `submit_form`, it's not possible to set arbitrary values for `RadioGroup` and `SelectElement` as well. You'd have to add or modify some child node in the dom first. That matches changing or dropping the maxlength attribute to be able to set an exceeding value or you can use `.set('value', ...`. If you want to construct some markup with a value attribute that exceeds a maxlength attribute for whatever reason, you can also use `.set('value', ...`. Overall, I don't see much demand to exceed a maxlength value though. In 99.9 % of all cases it probably does not make sense and actually should fail.

The reason behind this idea is that I have a system that handles authentication on multiple different websites and I often encounter cases where people provide a password that exceeds a given maxlength value and wonder why it does not work although it works in the browser. The explanation is that when they use their browser they first enter the first maxlength characters. The rendered input element is then full of obscured characters up to the right end. Then they continue typing but the browser does not actually accept those additional characters. However, what's displayed does not change and you cannot tell if the characters got added or not so that users do not even realize that only a prefix of the password they always type is actually used and expected.

To catch those cases, I'd have to add explicit checks everywhere, but as I said, I do think that such a check makes sense in 99.9 % of all use cases. That's why I submitted it here instead. I understand your reservation but I'm convinced that it does make sense at this place. Just think about it, make up your mind and give it a :+1: or :-1:.

cschramm:
Did you give it another thought? I'm fine with closing if you don't like it.

scoder:
I think this should be done by a separate tool.

PR #261, opened 2018-03-07
Add huge_tree support to HTMLParser.

stranac:
`HTMLParser` currently has no support for the `huge_tree` argument, which makes it impossible to correctly parse html documents with large text nodes.
This has been reported as an issue for Scrapy (https://github.com/scrapy/scrapy/issues/3077), and I was able to trace it back to lxml.


scoder:
Thanks!

PR #263, opened 2018-04-13
Add missing huge_tree parameter to XMLParser docstring

johnthagen:
The `huge_tree` parameter is missing from the docstring definition of the `XMLParser` class. This causes PyCharm to report a false positive when this parameter is used: https://youtrack.jetbrains.com/issue/PY-21959

This PR adds the parameter to the docstring.

Bug reported at: https://bugs.launchpad.net/lxml/+bug/1765437


scoder:
Thanks!

PR #264, opened 2018-05-05
Remove unused ez_setup.py

jwilk:
This module is unused since 78c92e59f5eb2a7e4323b2033eefc27f855a5ed6.


scoder:
Right, and it's hugely outdated as well. Thanks!

PR #265, opened 2018-05-14
Link libz only for static builds

nehaljwani:
If a user is not building libxml2 and is linking to it dynamically, then
there is no additional need to link to libz, since libxml2 will be
depending on libz anyway.


scoder:
I reverted this change for lxml 4.2.3, just released.

nehaljwani:
But why didn't the 4.2.2 wheels link to libz statically?

scoder:
Short answer: I don't know, and reverting the change was a much quicker solution than anything else.

PR #267, opened 2018-06-11
Let ElementTreeProducer use the available namespaces

regebro:
sax.ElementTreeProducer would ignore the namespace prefixes that were available in the element tree, and always generate new prefixes like ns00, ns01 etc.

This PR uses the prefixes in the XML by looking at the elements nsmap.


src/lxml/sax.py
```diff
@@ -165,6 +165,7 @@ def __init__(self, element_or_tree, content_handler):
         from xml.sax.xmlreader import AttributesNSImpl as attr_class
         self._attr_class = attr_class
         self._empty_attributes = attr_class({}, {})
+        self._prefix_map = dict((v, k) for (k, v) in element.nsmap.items())
```
scoder: I think it should check the prefix map of the element that it currently processes, not the root element.

scoder:
Tests are missing, should probably go into `test_sax.py`

src/lxml/sax.py
regebro: Sounds reasonable, I'll look into that.

src/lxml/sax.py
```diff
@@ -232,14 +235,25 @@ def _recursive_saxify(self, element, prefixes):
...
+
         try:
             prefix = prefixes[ns_uri]
         except KeyError:
-            prefix = prefixes[ns_uri] = 'ns%02d' % len(prefixes)
+            # New prefix
+            if element_prefixes and ns_uri in element_prefixes:
+                prefix = element_prefixes[ns_uri]
+            else:
+                # Unknown prefix, generate one
```
regebro: While writing tests I wasn't able to make a element tree that actually has an undefined namespace. I don't know if this code then is unnecessary, or if I'm not good enough at bending etree's to my will. :-)

src/lxml/sax.py
```diff
@@ -200,6 +200,7 @@ def _recursive_saxify(self, element, prefixes):
                 content_handler.characters(element.tail)
             return
 
+        element_prefixes = dict((v, k) for (k, v) in element.nsmap.items())
```
scoder: One problem here: if there is more than one prefix defined for a namespace URI, it becomes non-deterministic which one gets used. This is not as unusual as it might seem. For example, it's not uncommon to have a namespace as both the default and a prefixed namespace, e.g. when namespaced attributes are involved.

Another thing is that the nsmap is not cached internally but gets freshly created on each request. That's not hugely expensive (and we're already copying an entire tree through Python here anyway), but it would still add up since it's done for every element. Not sure if there is anything easy that could be done about it here.

And finally, while looking through the original code, I noticed something that looks like a bug. The `prefixes` mapping is collected globally, whereas it can actually differ between separate subtrees. If one subtree overrides a prefix that is defined further up, the current `ElementTreeProducer` could ignore that and start returning an incorrect prefix. I think the namespace handling should take advantage of the recursion to keep track of the currently valid prefix mapping. It has to do that anyway since it must report namespace SAX events.

Would you like to look into that problem? It might actually lead to an even better implementation.

src/lxml/sax.py
regebro: That code is unnecessary, I added a test and the ns0 is generated elsewhere.

src/lxml/sax.py
regebro: Sorry this took such long time, I ended up rewriting xmldiff from scratch, and not using the saxifier...

Anyway, I pushed another, better, attempt that should deal with the recursion. It also should prefer to not use a prefix for a default namespace. I don't have tests for that yet, will work on it shortly.

regebro: And I added a test, and now make the saxify prefer the prefix that the element actually had.

src/lxml/sax.py
```diff
@@ -202,7 +202,15 @@ def _recursive_saxify(self, element, prefixes):
...
             return
 
+        # Get a new copy in this call, so changes doesn't propagate upwards
+        prefixes = prefixes.copy()
         new_prefixes = []
+        for prefix, ns_uri in element.nsmap.items():
+            if prefixes.get(prefix) != ns_uri:
+                # New or updated namespace
+                new_prefixes.append( (prefix, ns_uri) )
+                prefixes[prefix] = ns_uri
```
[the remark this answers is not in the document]
regebro: Yeah, we still need to detect new namespace definitions, but it's probably better to compare the element.nsmap with the parents nsmap.

src/lxml/sax.py
```diff
@@ -212,13 +220,13 @@ def _recursive_saxify(self, element, prefixes):
...
                 attr_qnames[attr_ns_tuple] = build_qname(
-                    attr_ns_tuple[0], attr_ns_tuple[1], prefixes, new_prefixes)
+                    attr_ns_tuple[0], attr_ns_tuple[1], prefixes, None)
             sax_attributes = self._attr_class(attr_values, attr_qnames)
         else:
             sax_attributes = self._empty_attributes
 
         ns_uri, local_name = _getNsTag(tag)
-        qname = build_qname(ns_uri, local_name, prefixes, new_prefixes)
+        qname = build_qname(ns_uri, local_name, prefixes, element.prefix)
```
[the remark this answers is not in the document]
regebro: Except for attributes, yes.

src/lxml/sax.py
```diff
@@ -234,14 +242,19 @@ def _recursive_saxify(self, element, prefixes):
...
-            prefix = prefixes[ns_uri]
-        except KeyError:
-            prefix = prefixes[ns_uri] = 'ns%02d' % len(prefixes)
-            new_prefixes.append( (prefix, ns_uri) )
+
+        if preferred in prefixes and prefixes[preferred] == ns_uri:
+            prefix = preferred
+        else:
+            # Pick the first matching prefix
+            prefix = [pfx for pfx, uri in prefixes.items() if uri == ns_uri][0]
```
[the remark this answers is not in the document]
regebro: I'll just sort them, so it prefers them in alphabetical order.

src/lxml/sax.py
```diff
@@ -212,13 +220,13 @@ def _recursive_saxify(self, element, prefixes):
                 attr_ns_tuple = _getNsTag(attr_ns_name)
                 attr_values[attr_ns_tuple] = value
                 attr_qnames[attr_ns_tuple] = build_qname(
-                    attr_ns_tuple[0], attr_ns_tuple[1], prefixes, new_prefixes)
+                    attr_ns_tuple[0], attr_ns_tuple[1], prefixes, None)
```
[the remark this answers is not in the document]
regebro: I see. Tricky. I'll look into this.

regebro:
I'm not sure what happened with the tests here. I did, after much trouble replicate this locally, but I have no clue what is going on.

scoder:
I think that's the latest libxml2 release that strikes here, 2.9.9-rc1. I reported the problem on their side and configured the travis build (in the master branch) to use the previous release, which seems to work correctly. Rebasing on master should fix this. Sorry for the inconvenience.

src/lxml/sax.py
```diff
@@ -227,21 +235,35 @@ def _recursive_saxify(self, element, prefixes):
...
+
+        if prefixes.get(preferred_prefix) == ns_uri:
+            prefix = preferred_prefix
+        else:
+            # Pick the first matching prefix:
+            for pfx in sorted(prefixes, key=str):
+                if prefixes[pfx] == ns_uri:
+                    prefix = pfx
+                    if pfx is None and preferred_prefix == -1:
+                        # If preferred_prefix is -1, that's a flag to say
```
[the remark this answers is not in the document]
regebro: Yeah I know, I was just being lazy. :-) I'll change it.

src/lxml/sax.py
```diff
@@ -227,21 +235,35 @@ def _recursive_saxify(self, element, prefixes):
...
-            prefix = prefixes[ns_uri]
-        except KeyError:
-            prefix = prefixes[ns_uri] = 'ns%02d' % len(prefixes)
-            new_prefixes.append( (prefix, ns_uri) )
+
+        if prefixes.get(preferred_prefix) == ns_uri:
+            prefix = preferred_prefix
+        else:
+            # Pick the first matching prefix:
+            for pfx in sorted(prefixes, key=str):
```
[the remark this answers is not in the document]
regebro: I tested that and though namespaces ended up as always being lowercase? Maybe I messed up. Anyway, that is then an implicit and unclear feature, so that's not good code, I'll change this.

src/lxml/sax.py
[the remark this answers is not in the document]
regebro: Actually, that got ugly, I added a "is_attribute" parameter instead. See how you like that...

src/lxml/sax.py
```diff
@@ -235,37 +236,31 @@ def _recursive_saxify(self, element):
...
-                    prefix = pfx
-                    if pfx is None and preferred_prefix == -1:
-                        # If preferred_prefix is -1, that's a flag to say
-                        # that we want a prefix, any prefix, and only
-                        # accept the default prefix if no other is
-                        # available
-                        continue
-                    break
+            prefix = min(pfx for (pfx, uri) in nsmap.items()
+                         if pfx is not None and uri == ns_uri)
```
scoder: One more thing to think about: this will fail now if there is only the default prefix or if the URI does not have a prefix. (`min()` requires at least one value.) The second case probably shouldn't happen (and could be considered an actual error case), but the first seems ok and shouldn't fail.

scoder:
Thanks for your sustained work on this.

regebro:
No worries, it's been fun. Thanks for the help!

PR #268, opened 2018-07-30
Fixing possible memory corruption in doc dictionary if node is moved between docs

weggerle-a:
etree.insert function tries to handle the case when a node is moved
between documents with the function moveNodeToDocument. So far the
source_doc is taken from the destination child node which is wrong.
The moveNodeToDocument function will not fix the names in the
document dictionaries because source and target doc are the same.

The fix takes now the source_doc from the node element which
should be inserted.

This fixes issue https://bugs.launchpad.net/lxml/+bug/1773749

### DIFF
--- a/src/lxml/etree.pyx
+++ b/src/lxml/etree.pyx
@@ -917,7 +917,7 @@ cdef public class _Element [ type LxmlElementType, object LxmlElement ]:
         if c_node is NULL:
             _appendChild(self, element)
             return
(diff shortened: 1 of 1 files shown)


scoder:
Interesting bug. Thanks!

scoder:
Fix released in lxml 4.2.4.

weggerle-a:
Thanks for the fast merge and new release!

PR #269, opened 2018-08-09
Make .nsmap available in XSLT extensions.

adelton:
When adding namespaced elements/attributes in `execute` of XSLT extension element, new `xmlns` declaration is generated with prefix `ns0`. At the same time it does not seem to be possible to figure out what are the existing namespaces and prefixes in the `input_node`, resulting in
```
AttributeError: 'lxml.etree._ReadOnlyElementProxy' object has no attribute 'nsmap'
```
This patch makes it possible to see the namespace map.

### DIFF
--- a/src/lxml/readonlytree.pxi
+++ b/src/lxml/readonlytree.pxi
@@ -290,6 +290,26 @@ cdef class _ReadOnlyElementProxy(_ReadOnlyProxy):
                     return funicode(self._c_node.ns.prefix)
             return None
 
+    property nsmap:
+        u"""Namespace prefix->URI mapping known in the context of this
+        Element.
+        """
+        def __get__(self):
+            self._assertNode()
+            cdef xmlNode* c_node
+            cdef xmlNs* c_ns
+            nsmap = {}
+            c_node = self._c_node
+            while c_node is not NULL and c_node.type == tree.XML_ELEMENT_NODE:
+                c_ns = c_node.nsDef
+                while c_ns is not NULL:
+                    prefix = funicodeOrNone(c_ns.prefix)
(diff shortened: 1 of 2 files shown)


src/lxml/readonlytree.pxi
```diff
@@ -290,6 +290,26 @@ cdef class _ReadOnlyElementProxy(_ReadOnlyProxy):
                     return funicode(self._c_node.ns.prefix)
             return None
 
+    property nsmap:
```
scoder: Looks like a copy of `_Element.nsmap`, so the dict building should go into a shared helper function.

src/lxml/tests/test_xslt.py
```diff
@@ -1936,6 +1936,42 @@ def execute(self, context, self_node, input_node, output_parent):
...
+</xsl:stylesheet>
+""")
+        class MyExt(etree.XSLTExtension):
+            def execute(self, context, self_node, input_node, output_parent):
+                output_parent.text = str(input_node.nsmap)
+
+        extensions = { ('extns', 'show-nsmap') : MyExt() }
+
+        result = tree.xslt(style, extensions=extensions)
+        self.assertEqual(etree.tostring(result, pretty_print=True), """\
```
scoder: This needs to compare to a bytes literal to work in Py3.

scoder:
Documentation and docstring missing. Changelog entry would also be nice.

adelton:
I'm afraid I won't be able to address the comments in the review in foreseeable future. Would you prefer this pull request closed, or left open, for the off chance that someone else in the community has the need and time to polish this?

PR #272, opened 2018-09-02
Fix index out of range when <body> or <head> is missing

FVolral:
If a HTML file that doesn't contain a <body> or <head> part is parsed, 
it can lead to a index out of range error when you try 
`getattr(lxmlElement, "body")`

### DIFF
--- a/src/lxml/html/__init__.py
+++ b/src/lxml/html/__init__.py
@@ -287,7 +287,9 @@ class HtmlMixin(object):
         Return the <body> element.  Can be called from a child element
         to get the document's head.
         """
-        return self.xpath('//body|//x:body', namespaces={'x':XHTML_NAMESPACE})[0]
+        result = self.xpath('//body|//x:body', namespaces={'x':XHTML_NAMESPACE})
+        if len(result) > 1:
+            return result[0]
 
     @property
     def head(self):
@@ -295,7 +297,9 @@ class HtmlMixin(object):
         Returns the <head> element.  Can be called from a child
         element to get the document's head.
         """
-        return self.xpath('//head|//x:head', namespaces={'x':XHTML_NAMESPACE})[0]
+        result = self.xpath('//head|//x:head', namespaces={'x':XHTML_NAMESPACE})
+        if len(result) > 1:
+            return result[0]
 
     @property
     def label(self):


src/lxml/html/__init__.py
```diff
@@ -287,15 +287,19 @@ def body(self):
...
     @property
     def head(self):
         """
         Returns the <head> element.  Can be called from a child
         element to get the document's head.
         """
-        return self.xpath('//head|//x:head', namespaces={'x':XHTML_NAMESPACE})[0]
+        result = self.xpath('//head|//x:head', namespaces={'x':XHTML_NAMESPACE})
+        if len(result) > 1:
+            return result[0]
```
scoder: Otherwise … do what? Return None? Would be good to say so if that's what you want.

src/lxml/html/__init__.py
```diff
@@ -287,15 +287,19 @@ def body(self):
...
 
     @property
     def head(self):
         """
         Returns the <head> element.  Can be called from a child
         element to get the document's head.
         """
-        return self.xpath('//head|//x:head', namespaces={'x':XHTML_NAMESPACE})[0]
+        result = self.xpath('//head|//x:head', namespaces={'x':XHTML_NAMESPACE})
+        if len(result) > 1:
```
scoder: Why `> 1` ?

src/lxml/html/__init__.py
FVolral: I've some difficulty to run lxml tests, it's a bit new for me, and that pull request is a kind of experience.  It should return None indeed, I can correct that.

src/lxml/html/__init__.py
FVolral: "> 0",  because I'm stupid :p

src/lxml/html/__init__.py
charshack: None

charshack (approved):

Heartcore11 (approved):
src/lxml/html/__init__.py

src/lxml/html/__init__.py
Heartcore11: 11256de79711935479e9588aa26bc418cb8bb5ba

Heartcore11: > Otherwise … do what? Return None? Would be good to say so if that's what you want.

Heartcore11: > I've some difficulty to run lxml tests, it's a bit new for me, and that pull request is a kind of experience.  It should return None indeed, I can correct that.

Heartcore11: > 11256de79711935479e9588aa26bc418cb8bb5ba

https://github.com/lxml/lxml/pull/272#pullrequestreview-151623501

Heartcore11: > 11256de79711935479e9588aa26bc418cb8bb5ba

Heartcore11: > 11256de79711935479e9588aa26bc418cb8bb5ba

https://github.com/lxml/lxml

Heartcore11: > 11256de79711935479e9588aa26bc418cb8bb5ba

PR #274, opened 2018-10-01
Fix regression unknown error/undeclared vars while performing a xslt transform

weggerle-a:
Regression came up while updating from lxml 2.2.8. In a complex environment parameters were
suddenly reported as not declared after the lxml update like:
<xslt>:100:0:ERROR:XSLT:ERR_OK: runtime error, element 'result'
<string>:0:0:ERROR:XSLT:ERR_OK: Variable 'name' has not been declared.
<string>:0:0:ERROR:XSLT:ERR_OK: unknown error

Running the same transformation with the same template and inputs outside the
complex environment does not reproduce the error. The "complex environment" did several
xml operations before across multiple threads which seem to be important to reproduce the
issue.

At the end the issue seem to be caused by the string "name" which is contained in two dicts
and therefore with different addresses. The xsltXPathVariableLookup function does only a pointer
comparison and so "name" was != "name".

### DIFF
--- a/src/lxml/xslt.pxi
+++ b/src/lxml/xslt.pxi
@@ -526,16 +526,8 @@ cdef class XSLT:
         # anyway.
         if transform_ctxt.dict is not NULL:
             xmlparser.xmlDictFree(transform_ctxt.dict)
-        if kw:
-            # parameter values are stored in the dict
-            # => avoid unnecessarily cluttering the global dict
-            transform_ctxt.dict = xmlparser.xmlDictCreateSub(self._c_style.doc.dict)
-            if transform_ctxt.dict is NULL:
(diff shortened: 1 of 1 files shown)


scoder:
Thanks for bringing this up. However, I think it's not the right fix because, as the code comment suggests, it would store the parameter values in the global dict, and thus let it grow arbitrarily. I can't say right now how a better fix would look like, but it could be that the problem can be solved by calling the dict-name adaptation function at some point before the transformation, which is currently only run after the transformation to make sure all names come from the same dict.

weggerle-a:
I will try it out. Thanks for the suggestion. It's a pity that this kind of issues are so hard to reproduce...

scoder:
Closing, since the PR won't be accepted in its current form.

PR #277, opened 2018-11-14
use " double quotes in pretty printed XML declaration

eighthave:
While both ' and " are valid in XML,  by convention, " double quotes are standard in the XML Declaration.  Indeed lxml already uses " double quotes in the `DOCTYPE` declaration.

* https://en.wikipedia.org/wiki/XHTML#XML_declaration
* https://xmlwriter.net/xml_guide/xml_declaration.shtml
* https://www.tutorialspoint.com/xml/xml_declaration.htm
* https://developer.mozilla.org/en-US/docs/XML_introduction
* https://msdn.microsoft.com/en-us/library/ms256048(v=vs.110).aspx
* https://stackoverflow.com/questions/6800467/quotes-in-xml-single-or-double

For some context of where this can be important, pretty printing should create as few diffs as possible:
https://github.com/WeblateOrg/weblate/issues/2333

### DIFF
--- a/src/lxml/serializer.pxi
+++ b/src/lxml/serializer.pxi
@@ -273,16 +273,16 @@ cdef void _writeDeclarationToBuffer(tree.xmlOutputBuffer* c_buffer,
                                     int standalone) nogil:
     if version is NULL:
         version = <unsigned char*>"1.0"
-    tree.xmlOutputBufferWrite(c_buffer, 15, "<?xml version='")
+    tree.xmlOutputBufferWrite(c_buffer, 15, '''<?xml version="''')
     tree.xmlOutputBufferWriteString(c_buffer, <const_char*>version)
-    tree.xmlOutputBufferWrite(c_buffer, 12, "' encoding='")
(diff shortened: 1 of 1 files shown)


anthrotype:
isn't it a bit too late to change this?
FWIW cpython's elementtree also uses single quotes for the xml declaration

eighthave:
For all the uses I've had for lxml pretty printing, this change would be
helpful.  But yes, this is an old library, and there could be many uses
that would rather keep things as they are. IMHO, I'd also change
elementtree.

scoder:
I agree with @anthrotype. The fact that your change broke several tests in lxml's test suite (that you didn't seem to care about so far), suggests that it will introduce tons of problems on user side. And as you already said, there is no good reason for that change, since single and double quotes are treated identically by the XML spec.

eighthave:
I understand why you closed this, but for the record, there is a good
reason.  " double quotes are far and away the most common usage, so
using them minimizes diffs when using lxml pretty print mode.  lxml will
parse an XML file, then force the output to ' single quotes.

PR #281, opened 2019-06-03
Would it be possible to support Python on iOS?

cclauss:
Both of the major implementations of Python 3 on iOS have long standing issues about supporting lxml. omz/Pythonista-Issues#245 and ColdGrub1384/Pyto#25  However, Apple rejects their attempts to add lxml support because the AppStore guidelines prohibit accessing private APIs.  Is there a way for lxml to support iOS as a first class platform in such a way that lxml could be added without violating the AppStore guidelines?

Unrelated to the iOS questions above...  This PR proposes removing the Travis CI __sudo:__ tag because it has become a no-op now that sudo is _always_ available on Travis CI.

__sudo: required__ no longer is...  [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo:__ in your __.travis.yml__, we recommend removing that configuration_"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 os: linux
 dist: trusty
-sudo: false
 
 language: python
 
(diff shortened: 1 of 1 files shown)


scoder:
To answer the question: I have no experience with (or interest in) iOS myself, so, no support.
A static lxml build may work for this use case, but I don't know if that works out of the box or what it takes to make it work.

I'll merge the PR, though. Thanks.

cclauss:
Thanks for your (honest) answer and for landing the PR.

The default Travis distro is now xenial and it would be a good idea to test on bionic as well.
https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment
https://docs.travis-ci.com/user/reference/bionic/

@omz, @ColdGrub1384

PR #283, opened 2019-07-29
Fill python_requires with proper metadata

cedk:
Filling python_requires prevents pip from installing the package if it runs with an unsupported version.
See https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,8 @@ if versioninfo.is_pre_release():
 extra_options = {}
 if 'setuptools' in sys.modules:
     extra_options['zip_safe'] = False
+    extra_options['python_requires'] = (
+        '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, != 3.4.*')
 
     try:
         import pkg_resources
--- a/setup.py
+++ b/setup.py
@@ -227,7 +227,6 @@ an appropriate version of Cython installed.
     'Programming Language :: Python :: 2',
     'Programming Language :: Python :: 2.7',
     'Programming Language :: Python :: 3',
-    'Programming Language :: Python :: 3.4',
     'Programming Language :: Python :: 3.5',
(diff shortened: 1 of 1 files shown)


cedk:
Would it be possible to review and merge this quickly? We have our CI for old series which still support Python 3.4 that is failing because pip tries to install the version 4.4.0 instead of 4.3.5. And it will be great if 4.4.0 metadata could be fixed.

scoder:
Thanks. Not sure if it makes a difference, given that 4.0.0 still has the classifier.

scoder:
Note that you can instruct pip to install an older lxml version for a given Python version.
https://stackoverflow.com/questions/19559247/requirements-txt-depending-on-python-version

cedk:
Yes but it will not be clean. Our package should not care about which version of lxml is supported by which version of Python. It is the package manager job.
Indeed what will be great is to make a release 4.4.1 (or 4.4.0.post1) and delete release 4.4.0.

scoder:
> Yes but it will not be clean

But it solves the problem you are facing right now. And I doubt that there are still all that many users of Py3.4 who have not pinned their lxml version, now that Py3.4 is out of maintenance and does not receive security fixes any more.

cedk:
It is not nice to request every users to hard code the lxml version instead of using the proper way for upstream that the setuptools provides to describe which versions a package is compatible with.

scoder:
> request every users to hard code the lxml version

Honestly, every production use of any external library should pin the exact dependency version, or at most allow point version upgrades. Allowing arbitrary major version upgrades makes reproducible builds/deployments pretty much impossible.

cedk:
But I'm not talking about any specific production installation but for a public project. We can not decide for users to pin a specific version. What about bugs or security issues?
This is like for lxml and its extras_required, they are not pinned. What would happen if one of them decide to drop support of let says Python3.5 right now?

PR #285, opened 2019-08-02
Fixed broken link

servusDei2018:
The link labeled "doc/main.txt" leads to "lxml.de", which redirects to "https://www.fridaysforfuture.org/". This mistake is misleading, so I corrected the link to point to the real "doc/main.txt".


scoder:
Seems reasonable. Thanks.

PR #286, opened 2019-08-12
Fix: false detection of recursive include

RainerHausdorf:
In some cases ElementInclude does raise FatalIncludeError because of
recursive include detection. This is the case if the same file gets
included multiple times, but not recursive.

This is a fix for https://bugs.launchpad.net/lxml/+bug/1835708

### DIFF
--- a/src/lxml/ElementInclude.py
+++ b/src/lxml/ElementInclude.py
@@ -207,8 +207,7 @@ def _include(elem, loader=None, base_url=None,
                     raise FatalIncludeError(
                         "cannot load %r as %r" % (href, parse)
                         )
-                _parent_hrefs_for_node = set([href]) | _parent_hrefs
-                node = _include(node, loader, href, max_depth - 1, _parent_hrefs_for_node)
+                node = _include(node, loader, href, max_depth - 1, {href} | _parent_hrefs)
                 if e.tail:
                     node.tail = (node.tail or "") + e.tail
                 if parent is None:
--- a/src/lxml/ElementInclude.py
+++ b/src/lxml/ElementInclude.py
@@ -202,13 +202,13 @@ def _include(elem, loader=None, base_url=None,
                 if max_depth == 0:
                     raise LimitedRecursiveIncludeError(
                         "maximum xinclude depth reached when including file %s" % href)
-                _parent_hrefs.add(href)
                 node = load_include(href, parse, parser=parser)
                 if node is None:
                     raise FatalIncludeError(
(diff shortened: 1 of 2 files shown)


src/lxml/ElementInclude.py
```diff
@@ -202,13 +202,13 @@ def _include(elem, loader=None, base_url=None,
...
                     raise LimitedRecursiveIncludeError(
                         "maximum xinclude depth reached when including file %s" % href)
-                _parent_hrefs.add(href)
                 node = load_include(href, parse, parser=parser)
                 if node is None:
                     raise FatalIncludeError(
                         "cannot load %r as %r" % (href, parse)
                         )
-                node = _include(node, loader, href, max_depth - 1, _parent_hrefs)
+                _parent_hrefs_for_node = set([href]) | _parent_hrefs
```
scoder: Removing the `href` from the set after the recursion seems a less involved operation than creating a new list and two sets on each include. (Not that it makes a real difference in terms of time or space, but still …)

src/lxml/tests/test_etree.py
```diff
@@ -4389,6 +4420,16 @@ def test_xinclude_failures(self):
...
 
+    def test_multiple_include_of_same_file(self):
+        # Test that including the same file multiple times, but on the same level
+        # is not detected as recursive include
+        document = self.xinclude_loader("NonRecursive2.xml").getroottree()
+        self.include(document, self.xinclude_loader)
+
+        # same but for more than one level
+        document = self.xinclude_loader("NonRecursive1.xml").getroottree()
+        self.include(document, self.xinclude_loader)
```
scoder: I think another good test case would be to include the leaf file _only_ from the second level include files and not from the top-level file at all.

RainerHausdorf: added proposed testcase

src/lxml/ElementInclude.py
RainerHausdorf: Yes, I thought of that. But I think with the current implementation it is easier to understand (and guaranteed), that the call of `_include` is not able to change the parent_hrefs of the caller. But if you disagree, I can change it like you proposed.

scoder: I won't fight for it, but then please use at least a set literal.

RainerHausdorf: Good point! I did not know that this is faster.

scoder:
Thanks

PR #287, opened 2019-08-16
Domain expired/taken?

nkconnor:
the listed domain redirects to https://www.fridaysforfuture.org/


scoder:
All fine, works as intended.

PR #288, opened 2019-09-27
Disable using the space optimizing dictNames flag for HTML parsing contexts.

slovichon:
The SAX handlers perform buffer readahead accesses which result in
uninitialized data accesses because the HTML parsing API operates on
temporary buffers.

### DIFF
--- a/src/lxml/parser.pxi
+++ b/src/lxml/parser.pxi
@@ -127,7 +127,13 @@ cdef class _ParserDictionaryContext:
     cdef void initParserDict(self, xmlparser.xmlParserCtxt* pctxt):
         u"Assure we always use the same string dictionary."
         self.initThreadDictRef(&pctxt.dict)
-        pctxt.dictNames = 1
+        if not pctxt.html:
+            # dictNames cannot be enabled under HTML because the SAX
+            # handler performs buffer lookahead accesses for space
+            # optimization when it is enabled, but the HTML API operates
+            # on temp buffers which would result in accesses on
+            # uninitialized memory.
+          pctxt.dictNames = 1
 
     cdef void initXPathParserDict(self, xpath.xmlXPathContext* pctxt):
         u"Assure we always use the same string dictionary."
@@ -1197,8 +1203,6 @@ cdef void _initSaxDocument(void* ctxt) with gil:
 
     # set up document dict
     if c_doc and c_ctxt.dict and not c_doc.dict:
-        # I have no idea why libxml2 disables this - we need it
-        c_ctxt.dictNames = 1
         c_doc.dict = c_ctxt.dict
         xmlparser.xmlDictReference(c_ctxt.dict)
 


scoder:
Well … the comment that you removed says "we need it", so just disabling it this feature probably isn't the right solution. From your description, isn't this rather something to repair in libxml2?

slovichon:
I believe https://github.com/lxml/lxml/commit/7a48177fa904d00615b023dd654923eb4e0d40a6 was incorrect WRT setting dictNames = 1 unconditionally for the sax parser.  libxml gives html and xml parsing context initializers, and it sets dictNames = 1 for xml and 0 for html. 
 https://gitlab.gnome.org/GNOME/libxml2/blob/master/SAX2.c line 1880 accesses a lookahead read beyond the given buffer for space optimization, which is "OK" for XML, where the whole document buffer is available, but not for HTML, where temporary buffers are used; see https://gitlab.gnome.org/GNOME/libxml2/blob/master/HTMLparser.c line 2925.

I would like to see performance impact of using dictNames = 0 vs 1 when parsing html documents with lxml because those conditionals should rarely pass on uninitialized memory.

Arguably their API shouldn't be so brittle (i.e. operate on out of buffer bounds) but modifying internal parser options different than their initializers provide you with seems less correct, and that is the strong impression I was left with after filing an issue with them which was really the result of API misuse and not an inherent bug.

scoder:
> 7a48177
That code [was already there](
https://github.com/lxml/lxml/commit/7a48177fa904d00615b023dd654923eb4e0d40a6#diff-fe2acc82b7fb5bed0a3c950e40673ce0L236) before. And even the commit you reference is from 2013, and I never heard any complaints before this PR. With an out-of-bounds read access, I would expect a reproducible crash at least in some (however rare) cases for some input, but I'm not aware of any such example.

Note, I'm not saying that there is no problem here. I'm just trying to make sure we're not breaking lxml's internal assumptions to work around something that can be fixed elsewhere.

> SAX2.c line 1880
Are we looking at the same code? Both [this](https://gitlab.gnome.org/GNOME/libxml2/blob/64966ebefdbd350a2c34577f1bf88e39b3a48ece/SAX2.c#L1875) and [this](https://gitlab.gnome.org/GNOME/libxml2/blob/64966ebefdbd350a2c34577f1bf88e39b3a48ece/SAX2.c#L1927) look-ahead are guarded by `!ctxt->html`.

> after filing an issue with them
Could you point me to that ticket?

slovichon:
> > [7a48177](https://github.com/lxml/lxml/commit/7a48177fa904d00615b023dd654923eb4e0d40a6)
> 
> That code [was already there](https://github.com/lxml/lxml/commit/7a48177fa904d00615b023dd654923eb4e0d40a6#diff-fe2acc82b7fb5bed0a3c950e40673ce0L236) before.

Whoops, yes, I was wrong.  The source commit was https://github.com/lxml/lxml/commit/a47e07cd7415ee22346a91fd5b4deba14547afee.

> And even the commit you reference is from 2013, and I never heard any complaints before this PR. With an out-of-bounds read access, I would expect a reproducible crash at least in some (however rare) cases for some input, but I'm not aware of any such example.

So it may not be out of bounds per se, in terms of whether the memory is allocated, but there are definitely circumstances when it is accessing memory that is uninitialized.

You can see when unit testing with the following diff against HEAD:

```
Index: src/lxml/tests/test_htmlparser.py
===================================================================
--- src/lxml/tests/test_htmlparser.py
+++ src/lxml/tests/test_htmlparser.py
@@ -70,9 +70,18 @@ class HtmlParserTestCase(HelperTestCase)
         parser = self.etree.HTMLParser(recover=False)
         fromstring = self.etree.fromstring
         html = fromstring('''
-            <html><body id="bodyID"><p id="pID"></p></body></html>
+<html>
+<head>
+<style>
+.foo {margin: 1px;}
+</style>
+</head>
+<body>
+<div class="foo">{{ bar }}</div>
+</body>
+</html>
         ''', parser=parser)
-        self.assertEqual(len(html.xpath('//p[@id="pID"]')), 1)
+        self.assertEqual(len(html.xpath('//div[@class="foo"]')), 1)

     def test_html_ids_no_collect_ids(self):
         parser = self.etree.HTMLParser(recover=False, collect_ids=False)
```

then run ```valgrind python test.py htmlparser``` and see:

```
==56344== Conditional jump or move depends on uninitialised value(s)
==56344==    at 0x8122CA0: xmlSAX2TextNode (SAX2.c:1880)
==56344==    by 0x81258B4: xmlSAX2Characters (SAX2.c:2550)
==56344==    by 0x806EC8F: htmlParseScript (HTMLparser.c:2988)
==56344==    by 0x8071EBF: htmlParseContentInternal (HTMLparser.c:4616)
==56344==    by 0x8072460: htmlParseDocument__internal_alias.part.13 (HTMLparser.c:4813)
==56344==    by 0x8077582: htmlParseDocument__internal_alias (HTMLparser.c:4737)
==56344==    by 0x8077582: htmlDoRead (HTMLparser.c:6785)
==56344==    by 0x8077582: htmlCtxtReadMemory (HTMLparser.c:7071)
==56344==    by 0x7E08DD4: __pyx_f_4lxml_5etree_11_BaseParser__parseDoc (etree.c:117052)
==56344==    by 0x7E66B9D: __pyx_f_4lxml_5etree__parseDoc (etree.c:122832)
==56344==    by 0x7E67041: __pyx_f_4lxml_5etree__parseMemoryDocument (etree.c:124169)
```

(my line numbers aren't quite matching up but it should be obvious by looking at the libxml source code what the uninitialized accesses are originating from.)

> Note, I'm not saying that there is no problem here. I'm just trying to make sure we're not breaking lxml's internal assumptions to work around something that can be fixed elsewhere.

Yes, I was considering proposing to change the ```characters()``` API in libxml to pass an ```allow_readahead``` flag and disabling this where tmp buffer are used but specifically changing the flags after using the libxml initializer routine seemed like greater impropriety to address.

> > SAX2.c line 1880
> 
> Are we looking at the same code? Both [this](https://gitlab.gnome.org/GNOME/libxml2/blob/64966ebefdbd350a2c34577f1bf88e39b3a48ece/SAX2.c#L1875) and [this](https://gitlab.gnome.org/GNOME/libxml2/blob/64966ebefdbd350a2c34577f1bf88e39b3a48ece/SAX2.c#L1927) look-ahead are guarded by `!ctxt->html`.

?

Sorry, I do not see references to ```->html``` anywhere around SAX2.c:1875 and 1927...   the bad access is guarded by dictNames (L1880), which lxml is unconditionally enabling, disabling the guard.

> > after filing an issue with them
> Could you point me to that ticket?

It was more of pondering their API offering that led me here than opposition from libxml but the issue was at: https://gitlab.gnome.org/GNOME/libxml2/issues/106

scoder:
The read-ahead is happening here:
https://gitlab.gnome.org/GNOME/libxml2/blob/64966ebefdbd350a2c34577f1bf88e39b3a48ece/SAX2.c#L2070
It looks at `str[len]` (and, conditionally, `str[len + 1]`), which, for a string of length `len` is beyond the buffer end.

To fix this in libxml2, what I would try is either
a) changing the `COPY_BUF` macro in `HTMLparser.c` to always assign `\0` to the end of the copied buffer, so that we get a nul-terminated string or
b) test for `!ctxt->html` in `xmlSAX2TextNode()` right before looking at `cur` in both `else` cases.
I'd probably choose b).

I would also argue that the `XML_PARSE_COMPACT` block in `xmlSAX2TextNode()` is misplaced inside of the `dictNames` guard and should always be executed instead.

scoder:
This:
```
diff --git a/SAX2.c b/SAX2.c
index 5f141f9a..571489a4 100644
--- a/SAX2.c
+++ b/SAX2.c
@@ -1877,20 +1877,20 @@ xmlSAX2TextNode(xmlParserCtxtPtr ctxt, const xmlChar *str, int len) {
      * intern the formatting blanks found between tags, or the
      * very short strings
      */
-    if (ctxt->dictNames) {
-        xmlChar cur = str[len];
-
-	if ((len < (int) (2 * sizeof(void *))) &&
-	    (ctxt->options & XML_PARSE_COMPACT)) {
-	    /* store the string in the node overriding properties and nsDef */
-	    xmlChar *tmp = (xmlChar *) &(ret->properties);
-	    memcpy(tmp, str, len);
-	    tmp[len] = 0;
-	    intern = tmp;
-	} else if ((len <= 3) && ((cur == '"') || (cur == '\'') ||
+    if ((len < (int) (2 * sizeof(void *))) &&
+        (ctxt->options & XML_PARSE_COMPACT)) {
+        /* store the string in the node overriding properties and nsDef */
+        xmlChar *tmp = (xmlChar *) &(ret->properties);
+        memcpy(tmp, str, len);
+        tmp[len] = 0;
+        intern = tmp;
+    } else if (ctxt->dictNames) {
+	xmlChar cur = str[len];
+
+	} else if ((len <= 3) && (!cxtx->html) && ((cur == '"') || (cur == '\'') ||
 	    ((cur == '<') && (str[len + 1] != '!')))) {
 	    intern = xmlDictLookup(ctxt->dict, str, len);
-	} else if (IS_BLANK_CH(*str) && (len < 60) && (cur == '<') &&
+	} else if (IS_BLANK_CH(*str) && (len < 60) && (!cxtx->html) && (cur == '<') &&
 	           (str[len + 1] != '!')) {
 	    int i;
 
```

scoder:
(Sorry, wrong button)

slovichon:
Yes.  At first I only examined ```xmlSAX2TextNode()``` from the call site ```htmlParseScript()```, which seemed like a classic out of bounds access, but after asking for a test example I realized their unit tests use their API as expected, so I considered proposing to add a ```allow_readahead``` flag but figured addressing the 'API misuse' was better warranted as a next move.

slovichon:
Seemingly will be fixed in upstream libxml now, thanks for your attention with this problem.

PR #289, opened 2019-10-24
appveyor: test with Python 3.8

mr-c:


scoder:
Hmm, tried it, doesn't work for me:
https://ci.appveyor.com/project/St/lxml/builds/28411295/job/3rra0vdie7no3b1h
It builds a Py2.7 wheel instead of 3.8.

scoder:
I think it's because the `PYTHON` environment variable means different things in the appveyor matrix and the code snippet from which you copied the solution. I'll play around with it a bit to see if it's fixable.

t-mart:
Maybe you all did something in the meantime, but this now works for me on Windows.

(I was asking about 3.8 support at https://bugs.launchpad.net/lxml/+bug/1849424. I will now close this bug.)

Thanks for getting this out!

Please also make sure to update the setup.py with the 3.8 classifier:
```
Programming Language :: Python :: 3.8
```
and probably add 3.8 to the tox.ini:
```
envlist = py27, py34, py35, py36, py37, py38
```

mr-c:
Thanks @scoder I've made a follow up PR taking your comments & @t-mart 's into account at https://github.com/lxml/lxml/pull/290

@scoder Could "Rolling builds" be turned on in the Appveyor UI? It would speed uptesting. Basically each new commit to a PR would cancel Appveyor builds on prior commits for that PR: https://www.appveyor.com/docs/build-configuration/#rolling-builds

PR #290, opened 2019-10-28
appveyor: fix use of the selected Python version

mr-c:


anthrotype:
isn't line 17 (`SET PATH ...`) supposed to make sure that the selected python version is in the `%PATH%`?
Maybe the `%PYTHON%` enviornment variable isn't defined.

mr-c:
@anthrotype That is what I thought. This PR syncs with a known working configuration https://github.com/common-workflow-language/schema_salad/blob/master/appveyor.yml

mr-c:
> Now that you have defined `PYTHON` in the environment matrix and set `PATH` accordingly,

Well, now we do `set PATH=%PYTHON%\\Scripts;%PATH%` instead of `SET PATH=C:\\Python%PYTHON%;c:\\Python%PYTHON%\\scripts;%PATH%` so the full path is needed to `python.exe`

I agree about 
> `python -m pip` works (`pip.__main__` is unnecessary)

anthrotype:
> Well, now we do set PATH=%PYTHON%\\Scripts;%PATH% instead of

I see. But I not simply keeping `%PYTHON%` directory itself in the PATH and calling python as `python` as one would normally do?

mr-c:
@anthrotype I was sticking to a known good recipe. I am happy to make your change, if the `lxml` maintainers wish for it

scoder:
Superseded by the current build setup.

PR #294, opened 2019-12-03
make it easier to inspect element 

wxw-matt:
repr will show more information.
For example, for tag a, it will display something like this:
```
<Element a at 0x103c20820, {'href': '#', 'onclick': "javascript:windowHandle = window.open('http://example.com','popup', 'toolbar=0,scrollbars=1,location=0,statusbar=0,menubar=0,resizable=1,width=1020,height=600,left = 1,top = 1');windowHandle.focus();"}>
```
instead of
```
<Element a at 0x103c20820>.
```
The latter is not very useful.

### DIFF
--- a/src/lxml/etree.pyx
+++ b/src/lxml/etree.pyx
@@ -1121,7 +1121,10 @@ cdef public class _Element [ type LxmlElementType, object LxmlElement ]:
     # ACCESSORS
     def __repr__(self):
         u"__repr__(self)"
-        return "<Element %s at 0x%x, %s>" % (strrepr(self.tag), id(self), self.attrib)
+        if self.text:
+            return "<Element %s at 0x%x, %s(%s)>" % (strrepr(self.tag), id(self), self.attrib, self.text)
+        else:
+            return "<Element %s at 0x%x, %s>" % (strrepr(self.tag), id(self), self.attrib)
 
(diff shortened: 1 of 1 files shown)


scoder:
I'm unsure about this. It seems like a nice improvement, but it can also be very verbose. Just look at the `onclick` argument in your example. Adding the inner text seems even worse, since text content can be megabytes of data.

The text values could be shortened, obviously, but then, how much work, code and future maintenance cost is it worth investing for this?

I'm feeling inclined to leave the `repr()` as it is now.

wxw-matt:
I do not think that is verbose and annoying. Because if someone is under debug mode to inspect what the current element is and what attributes it has, that means he/she wants to have a good look at the data or attributes contained by the element. if there are a lot of attributes attached with the current elements and repr can not print all of them, the user has to inspect the attributes one by one. I think it is not user- friendly. That's my point. Thank you for your time. Hopefully, you think about it again.

scoder:
It's not just about "debug mode". I'm sure some users will not appreciate it when they are in the Python REPL and type `root.find(".//x)", and it blows up their console with a ton of unformatted data, rather than just saying "yes, it's there". The output of `repr()` is used in lots of places. It shouldn't be lengthy.

I'd rather add support for pprint() instead, if possible.

PR #296, opened 2019-12-29
Update documentation for external cython modules

dgreisen:
Changes needed to compile the example:
* add `include_dirs` directive
* change imports to `lxml.includes.*`
* fix `ElementDefaultClassLookup` typo

Happy to make modifications if any of this is incorrect.

### DIFF
--- a/doc/capi.txt
+++ b/doc/capi.txt
@@ -49,8 +49,14 @@ This is the easiest way of extending lxml at the C level.  A Cython_
 
     # My Cython extension
 
+    # directive pointing compiler to lxml header files;
+    # use ``aliases={"LXML_PACKAGE_DIR": LXML_PACKAGE_DIR}``
+    # argument to cythonize in setup.py to dynamically
+    # determine dir at compile time
+    # distutils: include_dirs = LXML_PACKAGE_DIR
+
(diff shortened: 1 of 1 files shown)


doc/capi.txt
```diff
@@ -49,8 +49,14 @@ This is the easiest way of extending lxml at the C level.  A Cython_
 
     # My Cython extension
 
+    # directive pointing compiler to lxml header files;
+    # use ``aliases={"LXML_PACKAGE_DIR": LXML_PACKAGE_DIR}``
```
scoder: Writing `LXML_PACKAGE_DIR` on both sides here seems more confusing than necessary. It's not very clear which refers to the `include_dirs` value below, and which the user needs to provide (and how). It should work to use `lxml.includes.__path__` for the alias value (assuming `lxml` is already installed at build time, which `setuptools` and `pip` can take care of).

dgreisen: addressed. On my machine, `lxml.__path__` works, while `lxml.includes.__path__` does not, so I updated as requested but kept it pointing to lxml root directory

scoder: Ah, right. The `api` C header files end up in that directory, so `lxml.__path__` is correct.

PR #298, opened 2020-03-02
Make iter() work with qnames

xmo-odoo:
"QName" is supposed to be usable anywhere a tag name is expected and iter() should take any number of tag names for filtering, but before this change passing a QName to iter() results in an exception.

Fixes https://bugs.launchpad.net/lxml/+bug/1865141


scoder:
Thanks!

PR #300, opened 2020-03-04
Simplify checks for minimum library versions

hughmcmaster:
This patch builds on the previous patch by simplifying the check for minimum library versions.

### DIFF
--- a/setupinfo.py
+++ b/setupinfo.py
@@ -110,17 +110,7 @@ def ext_modules(static_include_dirs, static_library_dirs,
         use_cython = False
         print("Building without Cython.")
 
-    lib_versions = get_library_versions()
-    versions_ok = True
-    if lib_versions[0]:
-        print("Using build configuration of libxml2 %s and libxslt %s" %
-              lib_versions)
-        versions_ok = check_min_version(lib_versions[0], (2, 7, 0), 'libxml2')
-    else:
-        print("Using build configuration of libxslt %s" %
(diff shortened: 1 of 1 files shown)


setupinfo.py
```diff
@@ -445,6 +434,17 @@ def get_library_versions():
     sys.exit(1)
 
 
+def check_build_dependencies():
+    xml2_version, xslt_version = get_library_versions()
+    if (not check_min_version(xml2_version, '2.7.0', 'libxml2') or
+            not check_min_version(xslt_version, '1.1.23', 'libxslt')):
+        sys.exit(1)
```
scoder: I know, this is part of a build script, but I'd rather not see this _function_ kill the runtime. A return value is always better for a function like this.

hughmcmaster: Doesn't that make it possible that the build might fail if a user's version of libxml2 and/or libxslt is too old? That said, it's probably unlikely, as those versions have been available for several years now.

Would you be happy with returning True/False, then from the caller, raising an exception? I'll push that version.

scoder: > from the caller, raising an exception

Yes, that's what I meant. There's probably a place higher up where raising such an exception would be expected.

hughmcmaster: Okay. I've updated the MR.

hughmcmaster:
Hi there, just checking in on this.

scoder:
Thanks

PR #302, opened 2020-06-27
Make mkhtml.py Python 3 compatible by replacing itervalues()

cjmayo:
https://docs.python.org/3/whatsnew/3.0.html#views-and-iterators-instead-of-lists

### DIFF
--- a/doc/mkhtml.py
+++ b/doc/mkhtml.py
@@ -272,7 +272,7 @@ def publish(dirname, lxml_path, release):
     SubElement(SubElement(menu_div[-1], 'li'), 'a', href='/sitemap.html').text = 'Sitemap'
 
     # integrate menu into web pages
-    for tree, basename, outpath in trees.itervalues():
+    for tree, basename, outpath in trees.values():
         head = find_head(tree)[0]
         SubElement(head, 'script', type='text/javascript').text = menu_js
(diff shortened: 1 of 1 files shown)


scoder:
Interesting. I didn't even know that I was still running it in Py2. Seems worth changing it in the `Makefile` then, too.

cjmayo:
> Interesting. I didn't even know that I was still running it in Py2. Seems worth changing it in the `Makefile` then, too.

To change website to:

```
website: inplace3
	PYTHONPATH=src:$(PYTHONPATH) $(PYTHON3) doc/mkhtml.py doc/html . ${LXMLVERSION}
```
Can't have an html3...? (for me /usr/bin/python is Python 3 which is how I found the problem!). But html has an inplace too?

But I think there is a bigger problem with apihtml in that epydoc is long dead and Python 2 only and removed from e.g. Ubuntu 20.04. At least the lack of epydoc just means apihtml fails without breaking the build.

scoder:
Oh well. Yeah, that's a bit of a bigger change then. :)

Maybe switching to Sphinx' autodoc is an option? I don't think it existed when I last looked for a replacement years ago, or at least wasn't en par with epydoc. Probably worth another try now.

In any case, this PR is ok as it is and can stand all by itself. Thanks.

PR #309, opened 2020-07-25
Use sphinx-apidoc to create API reference

cjmayo:
Epydoc is Python 2 only and unmaintained.

sphinx-apidoc is run before the build step, to avoid duplicate entries
being created.

---

This is an alternative to #308 that generates the rst files every time.


Makefile
```diff
@@ -105,34 +105,33 @@ ftest_build: build
...
-apihtml: inplace
-	rm -fr doc/html/api
-	@[ -x "`which epydoc`" ] \
-		&& (cd src && echo "Generating API docs ..." && \
-			PYTHONPATH=. epydoc -v --docformat "restructuredtext en" \
-			-o ../doc/html/api --exclude='[.]html[.]tests|[.]_' \
-			--exclude-introspect='[.]usedoctest' \
-			--name "lxml API" --url / lxml/) \
-		|| (echo "not generating epydoc API documentation")
+apidoc: clean docclean
```
scoder, line 108: I think this is what needs the `inplace` lxml build as a dependency, not the other targets, right?

cjmayo: `sphinx-apidoc` appears to only scan files, which is how I read:


> Warning
>
> sphinx-apidoc generates source files that use sphinx.ext.autodoc to document all found modules. If any modules have side effects on import, these will be executed by autodoc when sphinx-build is run.
>
>If you document scripts (as opposed to library modules), make sure their main routine is protected by a if `__name__ == '__main__'` condition.

https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html


and the build inplace causes sphinx-apidoc to create duplicate entries

e.g.

```diff
--- lxml.rst
+++ lxml.rst
@@ -18,11 +18,16 @@
 
    lxml.ElementInclude
    lxml._elementpath
+   lxml._elementpath
+   lxml.builder
    lxml.builder
    lxml.cssselect
    lxml.doctestcompare
    lxml.etree
+   lxml.etree
    lxml.objectify
+   lxml.objectify
+   lxml.sax
    lxml.sax
 
 Module contents
```

Looks like excluding .so files avoids this, by adding "*.so":

```
sphinx-apidoc -e -P -T -o doc/api src/lxml "*includes" "*tests" "*pyclasslookup.py" "*usedoctest.py" "*html/_html5builder.py" "*.so"
```

That's a possibility but still needs a separate apidoc target to support apidf as well. I'm not sure if it gains anything?

cjmayo: The duplication seems to be because there is a system copy of lxml installed. But without a system lxml, sphinx-apidoc appears to work without compiling lxml.

cjmayo:
Unfortunately I still get the same result with 61d135e.

I've tried `python -m sphinx.ext.apidoc` instead of `sphinx-apidoc`.

It is the modules that have the shared libraries that are duplicated.

doc/api/conf.py
```diff
@@ -0,0 +1,58 @@
+import os
+import sys
+sys.path.insert(0, os.path.abspath('../../src'))
+# For soupparser in lxml.html.ElementSoup module:
+sys.path.insert(1, os.path.abspath('../../src/lxml/html'))
```
scoder, line 5: That's a bug in the `ElementSoup` (legacy) module. Suggests that no-one has been using that essentially forever.
```suggestion
```

scoder:
Let's get this in and fix the rest later. Thanks a lot!

PR #310, opened 2020-08-11
Implement __len__ on InputGetter which is expected by FieldsDict

AidanWoolley:


src/lxml/html/__init__.py
```diff
@@ -1232,6 +1232,9 @@ def __iter__(self):
         ## FIXME: kind of dumb to turn a list into an iterator, only
         ## to have it likely turned back into a list again :(
         return iter(self._all_xpath(self.form))
+    
+    def __len__(self):
+        return len(self.all_xpath(self.form))
```
scoder, line 1237:
```suggestion
        return len(self._all_xpath(self.form))
```

scoder:
Right. There's an `__iter__`, there should be a `__len__`. Thanks!

PR #312, opened 2020-11-10
Removed unused Zope Public License from docs folder

wenovus:
I don't see the license being used anywhere in the repo, e.g. grep for "Zope" gives nothing, nor is the license mentioned in LICENSES.txt.

If I am wrong and this license is being used, it should be added LICENSES.txt so users know to which files it applies.


scoder:
I don't know what it related to, either. Let's remove it. Whatever code there was that it was referring to was probably removed and replaced during the last 15 years.

scoder:
Thanks

PR #313, opened 2021-01-05
adds initial fuzzer for oss-fuzz integration.

DavidKorczynski:
Hi,

Given the popularity of lxml I was thinking that it would be nice to set up continuous fuzzing of lxml, by way of OSS-Fuzz. In this PR: https://github.com/google/oss-fuzz/pull/4908 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate lxml. This includes developing initial fuzzers as well as integrating into OSS-Fuzz, however, it is preferable to have the fuzzers upstream so I included it in this PR - if you are happy with having the fuzzers here then I will remove them from the OSS-Fuzz repository.

Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects. Python support was recently provided and it can also fuzz native extensions, i.e. for memory corruption errors. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

If you would like to integrate, the only thing I need is as list of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. Notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

### DIFF
--- a/src/lxml/tests/fuzz_xml_parse.py
+++ b/src/lxml/tests/fuzz_xml_parse.py
@@ -1,0 +1,26 @@
+
+"""
+
+Fuzzes the lxml.etree.XML function with the Atheris fuzzer.
+
+The goal is to catch unhandled exceptions and potential 
+memory corruption issues in auto-generated code.
+
+"""
+
+import atheris
+import sys
+
+from lxml import etree as et
+
+def TestOneInput(data):
+  fdp = atheris.FuzzedDataProvider(data)
+
+  try:
+    root = et.XML(fdp.ConsumeUnicode(sys.maxsize))
+  except et.XMLSyntaxError:
+    None
+  return
+
+atheris.Setup(sys.argv, TestOneInput, enable_python_coverage=True)
+atheris.Fuzz()
--- a/src/lxml/tests/fuzz_xml_parse.py
+++ b/src/lxml/tests/fuzz_xml_parse.py
@@ -18,5 +18,6 @@ def test_etree_xml(data):
     pass
   return
 
-atheris.Setup(sys.argv, test_etree_xml, enable_python_coverage=True)
-atheris.Fuzz()
+if __name__ == "__main__":
+    atheris.Setup(sys.argv, test_etree_xml, enable_python_coverage=True)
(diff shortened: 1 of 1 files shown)


scoder:
Thanks. I wouldn't mind adding this.
How is this supposed to be run? It's not a unit test (and your naming and code format is rather … special).
I guess this can't simply run as part of the test suite, can it?

DavidKorczynski:
The fuzzer is meant to be run by the [Atheris](https://github.com/google/atheris) fuzzer.

Essentially, the following should be enough:
```
pip3 install atheris
python3 fuzz_xml_parse.py
```

It is not meant to be run as part of the test suite and OSS-Fuzz will be running continuously over time. However, we can integrate into the CI such that OSS-Fuzz will run the fuzzer when a PR is made for X amount of minutes.  

I tried to make the naming follow the test suite more - the original `TestOneInput` came because the function resembles a function `LLVMFuzzerTestOneInput` from the libFuzzer world. However, it should be more meaningful now?

scoder:
I'll give it a try. Thanks!

DavidKorczynski:
Thank you @scoder - which email could I use for the bug reports on the OSS-Fuzz repo?

scoder:
Hmm. I don't think it's in a state that is worth adding to OSS-fuzz right now. The parser is really the least interesting part of lxml when it comes to fuzzing, since it's not part of lxml but of libxml2. If you want to test the parser, call `xmllint` directly.

When I wrote "I'll give it a try", I should have added "when I have a use for it". Looking at https://github.com/google/oss-fuzz/pull/4908, it seems that setting it up as a `Makefile` target isn't exactly trivial, so that may not be now. I tried this in 37eae21e132241e67d05776447d7394c153e82f0 and it failed to compile for me with a linker setup error (trying to use the GCC linker instead of the clang linker). I'll probably also need some more of the setup that you added there, so I'll stop here and come back to it … later.

PR #314, opened 2021-01-19
Make it possible to control STATIC_* setup variables from the environ…

etanol:
When trying to build lxml with a statically linked libxml, the `setup.py` file needs to be edited or else successfully building lxml is impossible.

This PR tries to provide a way to give more control over the static build settings from environment variables.

For example, assume that libxml and libxslt have been statically built and installed under `/my/salsa` and we want to build lxml so that libxml and libxslt are **embedded** into the extension modules.  With the suggested changes to the `setup.py` file applied, it should be possible to do it with a single `pip` command:

```
STATIC=true \
LXML_STATIC_INCLUDE_DIRS="/my/salsa/include/libxml2" \
LXML_STATIC_LIBRARY_DIRS="/my/salsa/lib" \
LXML_STATIC_BINARIES="/my/salsa/lib/libxslt.a:/my/salsa/lib/libexslt.a:/my/salsa/lib/libxml2.a" \
WITH_XML2_CONFIG="/my/salsa/bin/xml2-config" \
WITH_XSLT_CONFIG="/my/salsa/bin/xslt-config" \
      pip install --no-binary=:all: lxml
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,10 +28,10 @@ import setupinfo
 def static_env_list(name, separator=None):
     return [x.strip() for x in os.environ.get(name, "").split(separator) if x.strip()]
 
-STATIC_INCLUDE_DIRS = static_env_list("LXML_STATIC_INCLUDE_DIRS", os.pathsep)
-STATIC_LIBRARY_DIRS = static_env_list("LXML_STATIC_LIBRARY_DIRS", os.pathsep)
+STATIC_INCLUDE_DIRS = static_env_list("LXML_STATIC_INCLUDE_DIRS", separator=os.pathsep)
+STATIC_LIBRARY_DIRS = static_env_list("LXML_STATIC_LIBRARY_DIRS", separator=os.pathsep)
 STATIC_CFLAGS = static_env_list("LXML_STATIC_CFLAGS")
-STATIC_BINARIES = static_env_list("LXML_STATIC_BINARIES", os.pathsep)
+STATIC_BINARIES = static_env_list("LXML_STATIC_BINARIES", separator=os.pathsep)
 
 # create lxml-version.h file
 versioninfo.create_version_h()
--- a/setup.py
+++ b/setup.py
@@ -25,10 +25,13 @@ import setupinfo
 # override these and pass --static for a static build. See
 # doc/build.txt for more information. If you do not pass --static
 # changing this will have no effect.
-STATIC_INCLUDE_DIRS = []
-STATIC_LIBRARY_DIRS = []
-STATIC_CFLAGS = []
-STATIC_BINARIES = []
+def static_env_list(name, separator=None):
+    return [x.strip() for x in os.environ.get(name, "").split(separator) if x.strip()]
+
+STATIC_INCLUDE_DIRS = static_env_list("LXML_STATIC_INCLUDE_DIRS", os.pathsep)
+STATIC_LIBRARY_DIRS = static_env_list("LXML_STATIC_LIBRARY_DIRS", os.pathsep)
+STATIC_CFLAGS = static_env_list("LXML_STATIC_CFLAGS")
+STATIC_BINARIES = static_env_list("LXML_STATIC_BINARIES", os.pathsep)
 
 # create lxml-version.h file
 versioninfo.create_version_h()


setup.py
```diff
@@ -25,10 +25,13 @@
 # override these and pass --static for a static build. See
 # doc/build.txt for more information. If you do not pass --static
 # changing this will have no effect.
-STATIC_INCLUDE_DIRS = []
-STATIC_LIBRARY_DIRS = []
+def static_env_list(name):
+    return list(filter(None, map(lambda x: x.strip(), os.environ.get(name, "").split(","))))
```
scoder, line 29: This is almost incomprehensible at first and second glance. Could you please turn this into a list comprehension?

setup.py
```diff
@@ -25,10 +25,13 @@
...
+
+STATIC_INCLUDE_DIRS = static_env_list("STATIC_INCLUDE_DIRS")
+STATIC_LIBRARY_DIRS = static_env_list("STATIC_LIBRARY_DIRS")
 STATIC_CFLAGS = []
-STATIC_BINARIES = []
+STATIC_BINARIES = static_env_list("STATIC_BINARIES")
```
scoder, line 34: Wouldn't it be better to make the env variable names lxml specific, i.e. `LXML_STATIC_BINARIES` etc.?
Otherwise, they could be accidentally picked up from somewhere else in a larger dependency build.

setup.py
```diff
@@ -25,10 +25,13 @@
 # override these and pass --static for a static build. See
 # doc/build.txt for more information. If you do not pass --static
 # changing this will have no effect.
-STATIC_INCLUDE_DIRS = []
-STATIC_LIBRARY_DIRS = []
-STATIC_CFLAGS = []
-STATIC_BINARIES = []
+def static_env_list(name):
+    return [x.strip() for x in os.environ.get(name, "").split("|") if x.strip()]
```
scoder, line 29:
```suggestion
    return [x.strip() for x in os.environ.get(name, "").split(",") if x.strip()]
```

etanol (commented):
I thought of using `|` instead of `,` because compiler _CFLAGS_ usually rely on the comma.  For example, when passing linker flags through the front end, e.g. `-Wl,-s`.

setup.py
etanol: I thought of using `|` instead of `,` because compiler _CFLAGS_ usually rely on the comma.  For example, when passing linker flags through the front end, e.g. `-Wl,-s`.

scoder: Ah, sorry, I considered that an accidental edit artefact.
So, add an additional `separator` argument to the function and make it `';'` for path values and spaces (i.e. `None`) for CFLAGS?

etanol: Sure.  Instead of `;`, would it be better to use [`os.pathsep`][1]?

[1]: https://docs.python.org/3.9/library/os.html#os.pathsep

scoder: Absolutely :)

setup.py
```diff
@@ -25,10 +25,13 @@
...
-STATIC_LIBRARY_DIRS = []
-STATIC_CFLAGS = []
-STATIC_BINARIES = []
+def static_env_list(name, separator=None):
+    return [x.strip() for x in os.environ.get(name, "").split(separator) if x.strip()]
+
+STATIC_INCLUDE_DIRS = static_env_list("LXML_STATIC_INCLUDE_DIRS", os.pathsep)
+STATIC_LIBRARY_DIRS = static_env_list("LXML_STATIC_LIBRARY_DIRS", os.pathsep)
+STATIC_CFLAGS = static_env_list("LXML_STATIC_CFLAGS")
+STATIC_BINARIES = static_env_list("LXML_STATIC_BINARIES", os.pathsep)
```
scoder, line 34:
```suggestion
def static_env_list(name, separator=None):
    return [x.strip() for x in os.environ.get(name, "").split(separator) if x.strip()]

STATIC_INCLUDE_DIRS = static_env_list("LXML_STATIC_INCLUDE_DIRS", separator=os.pathsep)
STATIC_LIBRARY_DIRS = static_env_list("LXML_STATIC_LIBRARY_DIRS", separator=os.pathsep)
STATIC_CFLAGS = static_env_list("LXML_STATIC_CFLAGS")
STATIC_BINARIES = static_env_list("LXML_STATIC_BINARIES", separator=os.pathsep)
```

scoder:
Thanks

PR #317, opened 2021-03-23
enabling access to per-entity system_url

Thirdegree:
Same idea as the other @properties on this type, I would like access to the system url for ENTITY ... SYSTEM ...


scoder:
Could you please add another test for a property that has no system URL, so that the `None` case is also covered?

Thirdegree:
No problem, done.

scoder:
Thanks

PR #318, opened 2021-05-15
Fix banner layout

bkline:
Tweak the CSS layout to avoid having the banner overlap with the tag line at some resolutions on the project page. See https://bugs.launchpad.net/lxml/+bug/1928553 (esp. attached screen shot). There may be more subtle tweaks which would make breakpoint-specific adjustments, but I went for this simpler modification, which seems to produce acceptable results at all resolutions.


scoder:
Looks good. Thanks.

PR #320, opened 2021-06-21
add link to Github for PyPi

andriyor:
Warehouse now uses the `project_urls` provided to display links in the sidebar, as well as including them in API responses to help automation tool find the source code for Pillow. For example, see Django's [setup.cfg](https://github.com/django/django/blob/master/setup.cfg) and [PyPI listing](https://pypi.org/project/Django/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 .tox
 .idea
+.vscode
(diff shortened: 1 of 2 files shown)


scoder:
Thanks

PR #321, opened 2021-07-08
Fix typo in iso_schematron_message.xsl

eltociear:
Implmentation -> Implementation


scoder:
Thanks, it's interesting that someone actually reads these. :)

However, these are externally maintained files that we only keep a copy of. I would like to avoid diverging from them. If there's ever a new release, we'd just overwrite these files. Fixing typos in them isn't really helpful in that process.

PR #322, opened 2021-07-29
_tofilelikeC14N: Always close output buffer

encukou:
Hello!
I'm reviewing output from a static analyzer, and found a piece of code that looks OK on closer inspection, but could be made more *obviously* correct.
I'm not very familiar with the code, but:

If `writer.error_log.__enter__` or `c14n.xmlC14NDocSaveTo` would
raise an exception, `c_buffer` would leak.
It seems that currently, neither of those can actually raise
(they're small tight `cdef` functions), but there's no
guarantee they'll remain exception-free in the future.

But there's one more thing that potentially could leak (at
least Cython generates an `unlikely` `goto` block for it):
the lookup of `__exit__` that happens at the start of the
`with` block.

---
There might be more like this as I trawl through static analyzer warnings and filter out false positives; are PRs like this helpful?

### DIFF
--- a/src/lxml/serializer.pxi
+++ b/src/lxml/serializer.pxi
@@ -862,15 +862,17 @@ cdef _tofilelikeC14N(f, _Element element, bint exclusive, bint with_comments,
         elif hasattr(f, 'write'):
             writer   = _FilelikeWriter(f, compression=compression)
             c_buffer = writer._createOutputBuffer(NULL)
-            with writer.error_log:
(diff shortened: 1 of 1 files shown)


scoder:
Right, there's no guarantee that the context manager itself won't raise. (`c14n.xmlC14NDocSaveTo()` is a plain C library function.)

Thanks, I'm happy to take such PRs.

PR #330, opened 2021-10-22
Add support for Python 3.10

hugovk:
Python 3.10 was released on 2021-10-04:

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955


bastimeyer:
I really don't like being annoying and having to ask this follow-up question, but is there a chance for the 4.6.4 release this week? The missing wheels on 4.6.3 are causing issues for users of our project, and lxml 4.6.4 is a blocking issue for our next release which needs to support py3.10 on all platforms.
Thanks

PR #332, opened 2021-11-19
Bugfixed Python 2to3 Regression in c14n2 serialization

khillman:
as discussed within https://mail.python.org/archives/list/lxml@python.org/thread/6ZFBHFOVHOS5GFDOAMPCT6HM5HZPWQ4Q/
this might not be the final fix but it gives a test which shows the problem with python2 and python3 if this fix is not in place.


scoder:
Thanks. Could you please move the tests to the `ETreeC14NTestCase` in `test_etree.py` ?

scoder:
I pushed a proposed fix. If you could remove your code changes and only add the tests, then we'd see if my changes make them work.

https://github.com/lxml/lxml/commit/8a9579c32782f3d59b73bcf3e7d2fb3b52b80956

khillman:
> I pushed a proposed fix. If you could remove your code changes and only add the tests, then we'd see if my changes make them work.
> 
> [8a9579c](https://github.com/lxml/lxml/commit/8a9579c32782f3d59b73bcf3e7d2fb3b52b80956)

Thank you for your proposal - I reverted the changes and added the tests to the place you mentioned.
Today I also had a look into several places where namespaces are handled and it seems to me that not only the default namespaces are in utf8/bytecode - but instead several places which are using [_getNsTag](https://github.com/lxml/lxml/search?q=_getNsTag) -- but we'll see ;) -- within iterparse in contrast the namespaces are explicitely converted to unicode (see https://github.com/lxml/lxml/blob/master/src/lxml/iterparse.pxi#L433) and also the register_namespace converts them to bytestrings (https://github.com/lxml/lxml/blob/master/src/lxml/etree.pyx#L195) -- therefore I was a bit confused about how to fix this current issue the right way - whether to convert/use everything as unicode vs everything as utf8 bytecode ;)

scoder:
Looks like your tests do work now.

scoder:
> Today I also had a look into several places where namespaces are handled and it seems to me that not only the default namespaces are in utf8/bytecode - but instead several places which are using [_getNsTag](https://github.com/lxml/lxml/search?q=_getNsTag)

Yes, that's normal. Since the internal text representation in libxml2 is UTF-8, it's much more efficient to compare byte strings internally than to first decode it into a Unicode string copy and then compare that. For the C41N implementation (which is mostly the same as in CPython's ElemenTree), this isn't the case. Some places have a low-level implementation whereas more complex code often stays at the Python level. That's the cool thing about Cython, you can mix both freely in the same file. You just have to remember which assumptions you're making. :)

khillman:
> Looks like your tests do work now.

Yes, it looks good so far - so if all succeed you could merge the only the tests into the master to prevent future regressions regarding this issue.

PR #333, opened 2021-11-26
Use the non-depcrecated TextTestResult instead of _TextTestResult

hroncok:
_TextTestResult was removed from Python 3.11.
TextTestResult is available on all supported Python versions.


scoder:
Thanks

PR #335, opened 2022-01-12
Added note to documentation about XSLT bug

trevor87:


scoder:
Thanks. I'd rather move it down to the end of that section, though, after the paragraph that describes the supported features. This is (practically) an unsupported feature, so it chains nicely.

trevor87:
I moved the note to the end. As I do not have my devtools on this computer, I had to do it in Github and therefore could not test building the documentation again.

scoder:
Thanks

PR #337, opened 2022-01-27
Allow Pathlike objects for file arguments

janssenhenning:
See https://bugs.launchpad.net/lxml/+bug/1820837

### DIFF
--- a/src/lxml/apihelpers.pxi
+++ b/src/lxml/apihelpers.pxi
@@ -1585,6 +1585,9 @@ cdef bint _isFilePath(const_xmlChar* c_path):
 cdef object _encodeFilename(object filename):
     u"""Make sure a filename is 8-bit encoded (or None).
     """
+    if hasattr(filename, u"__fspath__") and python.PY_VERSION_HEX >= 0x03060000:
+        filename = os_fspath(filename)
+
     if filename is None:
         return None
     elif isinstance(filename, bytes):
@@ -1599,7 +1602,7 @@ cdef object _encodeFilename(object filename):
                 pass
         return filename8
     else:
-        raise TypeError("Argument must be string or unicode.")
+        raise TypeError("Argument must be string, unicode or PathLike.")
 
 cdef object _decodeFilename(const_xmlChar* c_path):
     u"""Make the filename a unicode string if we are in Py3.
--- a/src/lxml/includes/etree_defs.h
+++ b/src/lxml/includes/etree_defs.h
@@ -247,6 +247,22 @@ long _ftol2( double dblSource ) { return _ftol( dblSource ); }
 #define _isString(obj)   (PyUnicode_Check(obj) || PyBytes_Check(obj))
 #endif
 
+#if PY_MAJOR_VERSION < 3
+#define _isStringOrPathLike(obj)   (PyString_CheckExact(obj)  || \
+                                    PyUnicode_CheckExact(obj) || \
+                                    PyType_IsSubtype(Py_TYPE(obj), &PyBaseString_Type))
+#else
+#if PY_MINOR_VERSION < 6
+/* builtin subtype type checks are almost as fast as exact checks in Py2.7+
+ * and Unicode is more common in Py3 */
+#define _isStringOrPathLike(obj)   (PyUnicode_Check(obj) || PyBytes_Check(obj))
+#else
+#define _isStringOrPathLike(obj)   (PyUnicode_Check(obj) || \
+                                    PyBytes_Check(obj)   || \
+                                    PyObject_HasAttrString(obj, "__fspath__"))
+#endif
+#endif
+
 #define _isElement(c_node) \
         (((c_node)->type == XML_ELEMENT_NODE) || \
          ((c_node)->type == XML_COMMENT_NODE) || \
--- a/src/lxml/python.pxd
+++ b/src/lxml/python.pxd
@@ -121,6 +121,7 @@ cdef extern from "includes/etree_defs.h": # redefines some functions as macros
     cdef void lxml_free(void* mem)
     cdef void* lxml_unpack_xmldoc_capsule(object capsule, bint* is_owned) except? NULL
     cdef bint _isString(object obj)
+    cdef bint _isStringOrPathLike(object obj)
     cdef const_char* _fqtypename(object t)
     cdef object PY_NEW(object t)
     cdef bint LXML_UNICODE_STRINGS
--- a/src/lxml/tests/test_dtd.py
+++ b/src/lxml/tests/test_dtd.py
@@ -24,6 +24,25 @@ class ETreeDtdTestCase(HelperTestCase):
 
         dtd = etree.DTD(fileInTestDir("test.dtd"))
         self.assertTrue(dtd.validate(root))
+    
+    def test_dtd_file_pathlike(self):
+        parse = etree.parse
+        tree = parse(fileInTestDir("test.xml"))
+        root = tree.getroot()
+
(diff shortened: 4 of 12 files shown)


src/lxml/apihelpers.pxi
```diff
@@ -1585,6 +1585,9 @@ cdef bint _isFilePath(const_xmlChar* c_path):
 cdef object _encodeFilename(object filename):
     u"""Make sure a filename is 8-bit encoded (or None).
     """
+    if hasattr(filename, u"__fspath__") and python.PY_VERSION_HEX >= 0x03060000:
```
scoder, line 1588: I'd do the compile time check before the costly runtime check.
```suggestion
    if  python.PY_VERSION_HEX >= 0x03060000 and hasattr(filename, u"__fspath__"):
```

src/lxml/includes/etree_defs.h
```diff
@@ -247,6 +247,22 @@ long _ftol2( double dblSource ) { return _ftol( dblSource ); }
 #define _isString(obj)   (PyUnicode_Check(obj) || PyBytes_Check(obj))
 #endif
 
+#if PY_MAJOR_VERSION < 3
+#define _isStringOrPathLike(obj)   (PyString_CheckExact(obj)  || \
+                                    PyUnicode_CheckExact(obj) || \
+                                    PyType_IsSubtype(Py_TYPE(obj), &PyBaseString_Type))
+#else
+#if PY_MINOR_VERSION < 6
```
scoder, line 255: This is going to fail when Python 4.0 comes out. Use `PY_VERSION_HEX  < 0x03060000` instead.
Also, this seems to repeat what `_isString()` already does. Just use that.

src/lxml/includes/etree_defs.h
```diff
@@ -247,6 +247,22 @@ long _ftol2( double dblSource ) { return _ftol( dblSource ); }
...
+/* builtin subtype type checks are almost as fast as exact checks in Py2.7+
+ * and Unicode is more common in Py3 */
+#define _isStringOrPathLike(obj)   (PyUnicode_Check(obj) || PyBytes_Check(obj))
+#else
+#define _isStringOrPathLike(obj)   (PyUnicode_Check(obj) || \
+                                    PyBytes_Check(obj)   || \
+                                    PyObject_HasAttrString(obj, "__fspath__"))
```
scoder, line 262: This is much better implemented in Cython with `hasattr()` in order to avoid creating a new string object on each attribute test. And, in fact, I'd rather have a Cython function that extracts the path directly if there is one, rather than testing for the attribute, then testing for the attribute again and then calling `os.fspath()` (which tests for the attribute again).

Additionally, if we avoid calling `os.fspath()`, then we can make also it work in Python < 3.6. There's a [`PyOS_FSPath()`](https://docs.python.org/3/c-api/sys.html#c.PyOS_FSPath) that we can call in Py3.6+, and you can implement a suitable fallback here (or in Cython) that just reads `obj.__fspath__`. One way to do it is to define `PyOS_FSPath` (or rather a safe alias) to `NULL` here for Py<3.6 and to avoid calling it based on `PY_VERSION_HEX` from Cython, using a separate simple Cython implementation in that case.

scoder (commented):
Cool, very nice. I think we should backport it, though. There is no reason why this couldn't work in Py<3.6 if someone cares to define `__fspath__` on some object, such as a backported pathlib.

src/lxml/includes/etree_defs.h
janssenhenning: I hope I understood your suggestions correctly. I now removed the `_isStringOrPathLike` function and added a cython function `_getFSPathOrObject`, which uses `PyOS_FSPath()` for python 3.6+ and gets `__fspath__()` for older python versions. This function I put before the `_isString` checks for the file arguments

src/lxml/apihelpers.pxi
```diff
@@ -1582,6 +1582,25 @@ cdef bint _isFilePath(const_xmlChar* c_path):
...
+            fspath = python.PY_FSPath(obj)
+        except TypeError:
+            return obj
+        return fspath
+    elif hasattr(obj, '__fspath__'):
+        fspath = obj.__fspath__
+        if callable(fspath):
+            return fspath()
+        raise ValueError('__fspath__ is not callable')
+    return obj
```
scoder, line 1602: I think we can keep the `_isString()` check, because it's so incredibly cheap.
For Py<3.6, since users probably weren't expecting `__fspath__` to become special, I think ignoring anything we find that's not callable is better than shouting at them.
Finally, I consider `hasattr()` followed by an attribute access an anti-pattern. Why look up the attribute twice? If you want to read the attribute, do so. We can use the 3-args `getattr()` in this case, to avoid the likely `AttributeError` if possible.
```suggestion
cdef object _NO_FSPATH = object()

cdef object _getFSPathOrObject(object obj):
    """
    Get the __fspath__ attribute of an object if it exists.
    Otherwise, the original object is returned.
    """
    if _isString(obj):
        return obj
    if python.PY_VERSION_HEX >= 0x03060000:
        try:
            return python.PY_FSPath(obj)
        except TypeError:
            return obj
    fspath = getattr(obj, '__fspath__', _NO_FSPATH)
    if fspath is not _NO_FSPATH and callable(fspath):
        return fspath()
    return obj
```

src/lxml/tests/test_dtd.py
```diff
@@ -24,6 +24,20 @@ def test_dtd_file(self):
...
+    def test_dtd_file_pathlike(self):
+        parse = etree.parse
+        tree = parse(fileInTestDir("test.xml"))
+        root = tree.getroot()
+
+        class Path(object):
+            def __init__(self,path):
+                self.path = path
+            def __fspath__(self):
+                return self.path
```
scoder, line 37: There is a `common_imports.py` helper module for tests. You can just move this class there, so that you don't have to repeat it in all test methods. Maybe name it `SimpleFSPath` then, or so, to distinguish it from `pathlib`.

src/lxml/includes/etree_defs.h
```diff
@@ -247,6 +247,12 @@ long _ftol2( double dblSource ) { return _ftol( dblSource ); }
 #define _isString(obj)   (PyUnicode_Check(obj) || PyBytes_Check(obj))
 #endif
 
+#if PY_VERSION_HEX >= 0x03060000
+#define PY_FSPath(obj) (PyOS_FSPath(obj))
+#else
+#define PY_FSPath(obj) (NULL)
+#endif
```
scoder, line 254: This is a public header file, so we should avoid introducing user side collisions where possible.
```suggestion
#if PY_VERSION_HEX >= 0x03060000
#define lxml_PyOS_FSPath(obj) (PyOS_FSPath(obj))
#else
#define lxml_PyOS_FSPath(obj) (NULL)
#endif
```

src/lxml/python.pxd
```diff
@@ -127,6 +127,7 @@ cdef extern from "includes/etree_defs.h": # redefines some functions as macros
     cdef bint IS_PYTHON2
     cdef bint IS_PYTHON3  # legacy, avoid
     cdef bint IS_PYPY
+    cdef object PY_FSPath(object obj)
```
scoder, line 130:
```suggestion
    cdef object PY_FSPath "lxml_PyOS_FSPath" (object obj)
```

src/lxml/tests/test_dtd.py
janssenhenning: Done.

scoder:
Thanks!

PR #338, opened 2022-02-12
Respect case of xsi:double infinity and NaN

haxtibal:
See [lxml launchpad bug 1960715](https://bugs.launchpad.net/lxml/+bug/1960715).

I'm not too experienced with Cython. Feel free to edit the commit.

### DIFF
--- a/src/lxml/objectify.pyx
+++ b/src/lxml/objectify.pyx
@@ -38,6 +38,10 @@ import_lxml__etree()
 
 __version__ = etree.__version__
 
+import math
+cdef object _float_is_inf = math.isinf
+cdef object _float_is_nan = math.isnan
+
 cdef object re
 import re
 
@@ -1205,8 +1209,17 @@ cdef dict _PYTYPE_DICT = {}
 cdef dict _SCHEMA_TYPE_DICT = {}
 cdef list _TYPE_CHECKS = []
 
-cdef unicode _lower_bool(b):
-    return u"true" if b else u"false"
+cdef unicode _xml_bool(value):
+    return u"true" if value else u"false"
+
+cdef unicode _xml_float(value):
+    if _float_is_inf(value):
+        if value > 0:
+            return u"INF"
+        return u"-INF"
+    if _float_is_nan(value):
+        return u"NaN"
+    return unicode(repr(value))
 
 cdef _pytypename(obj):
     return u"str" if python._isString(obj) else _typename(obj)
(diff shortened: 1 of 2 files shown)


src/lxml/objectify.pyx
```diff
@@ -1205,8 +1208,17 @@ cdef dict _PYTYPE_DICT = {}
 cdef dict _SCHEMA_TYPE_DICT = {}
 cdef list _TYPE_CHECKS = []
 
-cdef unicode _lower_bool(b):
-    return u"true" if b else u"false"
+cdef unicode _lower_bool(value):
```
haxtibal, line 1211: Btw., changing arg name here is unrelated, but was required because of cython/cython#4019.

scoder:
Thanks. Could you please add tests in `src/lxml/tests/test_objectify.py`? It looks like we only test these values as input, not as output or type markers.

src/lxml/objectify.pyx
```diff
@@ -1205,8 +1208,17 @@ cdef dict _PYTYPE_DICT = {}
...
+    return u"true" if value else u"false"
+
+cdef unicode _case_sensitive_float(value):
+    if math.isinf(value):
+        if value > 0:
+            return u"INF"
+        return u"-INF"
+    if math.isnan(value):
+        return u"NaN"
```
scoder, line 1220: Just a suggestion, but instead of `math`, you can import `isinf` and `isnan` directly as global `cdef` variables. Maybe rename them to `_float_is_inf` and `_float_is_nan`.

haxtibal: Did what I believe you mean. Honestly I can't yet fully reason about putting `cdef object` along with `import` vs. omitting the `cdef`. I see there's an additional runtime `if` in the C code for the latter case, but don't know when to do what. If you find a minute, would you explain it to me?

haxtibal:
> Could you please add tests in `src/lxml/tests/test_objectify.py`?

Added _some_ tests. Not sure if they're complete enough or in the right place, would you have a look?

src/lxml/objectify.pyx
scoder: Sure. Defining a global variable with `cdef` makes it a static C variable. Otherwise, it's a Python variable that gets looked up in a dict. That's much slower. Also, `cdef` variables are not exposed to the module dict, i.e. not visible to users.

src/lxml/objectify.pyx
```diff
@@ -38,6 +38,10 @@ import_lxml__etree()
 
 __version__ = etree.__version__
 
+import math
+cdef object _float_is_inf = math.isinf
+cdef object _float_is_nan = math.isnan
```
scoder, line 43: This puts `math` into the module dict. Let's avoid that.
```suggestion
cdef object _float_is_inf, _float_is_nan
from math import isinf as _float_is_inf, isnan as _float_is_nan
```

scoder (commented):
That looks good.

haxtibal:
Thanks for your review.

Recent force push was merely to align author and committer email in my commit. They differed as a result of bad git project settings on my side.

PR #340, opened 2022-03-07
Fix inheritance in lxml.html

xmo-odoo:
As the old comment / FIXME from 8132c755adad4a75ba855d985dd257493bccc7fd notes, the mixin should come first for the inheritance to be correct (the left-most class is the first in the MRO, at least if no diamond inheritance is involved).

Also fix the odd `super` in `HtmlMixin` likely stemming from the incorrect MRO.

Fixes the inheritance order of all `HTML*` base classes though it probably doesn't matter for other than `HtmlElement`.

### DIFF
--- a/src/lxml/html/tests/test_basic.py
+++ b/src/lxml/html/tests/test_basic.py
@@ -1,11 +1,51 @@
+import sys
 import unittest
 from lxml.tests.common_imports import make_doctest, doctest
-import lxml.html
+from lxml import html
+
+class TestBasicFeatures(unittest.TestCase):
+    def test_various_mixins(self):
+        base_url = "http://example.org"
+        doc = html.fromstring("""
+        <root>
+            <!-- comment -->
+            <?pi contents ?>
+            &entity;
+            <el/>
+        </root>
+        """, base_url=base_url)
+        self.assertEqual(doc.getroottree().docinfo.URL, base_url)
+        self.assertEqual(len(doc), 3)
+        self.assertIsInstance(doc[0], html.HtmlComment)
+        self.assertIsInstance(doc[1], html.HtmlProcessingInstruction)
+        self.assertIsInstance(doc[2], html.HtmlElement)
+        for child in doc:
+            # base_url makes sense on all nodes (kinda) whereas `classes` or
+            # `get_rel_links` not really
+            self.assertEqual(child.base_url, base_url)
+
+    def test_set_extension(self):
+        e = html.Element('e')
+        e.set('a')
+        e.set('b', None)
+        e.set('c', '')
+        self.assertEqual(
+            html.tostring(e),
+            b'<e a b c=""></e>',
+            "Attributes set to `None` should yield empty attributes"
+        )
+        self.assertEqual(e.get('a'), '', "getting the empty attribute results in an empty string")
+        self.assertEqual(e.attrib, {
+            'a': '',
+            'b': '',
+            'c': '',
+        })
 
 def test_suite():
     suite = unittest.TestSuite()
     suite.addTests([make_doctest('test_basic.txt')])
-    suite.addTests([doctest.DocTestSuite(lxml.html)])
+    suite.addTests([doctest.DocTestSuite(html)])
+    suite.addTest(unittest.TestLoader().loadTestsFromModule(sys.modules[__name__]))
     return suite
 
 if __name__ == '__main__':
--- a/src/lxml/html/tests/test_basic.py
+++ b/src/lxml/html/tests/test_basic.py
@@ -24,7 +24,7 @@ class TestBasicFeatures(unittest.TestCase):
             # `get_rel_links` not really
             self.assertEqual(child.base_url, base_url)
 
-    def test_set_extension(self):
+    def test_set_empty_attribute(self):
         e = html.Element('e')
         e.set('a')
         e.set('b', None)
--- a/src/lxml/html/__init__.py
+++ b/src/lxml/html/__init__.py
@@ -245,7 +245,7 @@ class HtmlMixin(object):
         creates a 'boolean' attribute without value, e.g. "<form novalidate></form>"
         for ``form.set('novalidate')``.
         """
-        super(HtmlElement, self).set(key, value)
+        super(HtmlMixin, self).set(key, value)
 
     @property
     def classes(self):
@@ -685,21 +685,19 @@ iterlinks = _MethodFunc('iterlinks', copy=False)
 rewrite_links = _MethodFunc('rewrite_links', copy=True)
 
 
-class HtmlComment(etree.CommentBase, HtmlMixin):
+class HtmlComment(HtmlMixin, etree.CommentBase):
     pass
 
 
-class HtmlElement(etree.ElementBase, HtmlMixin):
-    # Override etree.ElementBase.cssselect() and set(), despite the MRO (FIXME: change base order?)
-    cssselect = HtmlMixin.cssselect
-    set = HtmlMixin.set
-
-
-class HtmlProcessingInstruction(etree.PIBase, HtmlMixin):
+class HtmlElement(HtmlMixin, etree.ElementBase):
     pass
 
 
-class HtmlEntity(etree.EntityBase, HtmlMixin):
+class HtmlProcessingInstruction(HtmlMixin, etree.PIBase):
+    pass
+
+
+class HtmlEntity(HtmlMixin, etree.EntityBase):
     pass
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,7 @@
 envlist = py27, py35, py36, py37, py38, py39, py310
 
 [testenv]
+allowlist_externals = make
 setenv =
     CFLAGS = -g -O0
 commands =


xmo-odoo:
Also didn't find a contributing guide so not sure how forward- or back- ports might work (if they're done / supported for minor issues).

scoder:
Thanks. That looks correct. Can't say why it wasn't done like this at the time.

It's probably also missing tests for comments, PIs and entities, that's why it didn't show so far. Could you please add some?

> the *-mixins will only apply to the `classes` (or default fallbacks), if some element classes are created solely through mixins (which seems feasible) those won't get *-mixins, and the same applies to `HtmlElement` fallbacks (so *-mixin can't be used to apply changes to all elements from that lookup object).

Not sure what exactly you're referring to here. Do you mean elements created by instantiating element classes (rather than parsing)? That behaviour is documented on the [element classes page](https://lxml.de/element_classes.html).

Regarding backports, lxml tends to be mostly backwards compatible, so that rarely happens.

xmo-odoo:
> > the *-mixins will only apply to the `classes` (or default fallbacks), if some element classes are created solely through mixins (which seems feasible) those won't get *-mixins, and the same applies to `HtmlElement` fallbacks (so *-mixin can't be used to apply changes to all elements from that lookup object).
> 
> Not sure what exactly you're referring to here. Do you mean elements created by instantiating element classes (rather than parsing)? That behaviour is documented on the [element classes page](https://lxml.de/element_classes.html).

No I mean that if we set up a parser with the lookup
```python
HtmlElementClassLookup(mixins=[
    ('*', A),
    ('foo', B),
])
```
Then parsed `foo` elements will not get the `A` mixin, at least according to my reading (as noted I haven't tested it yet, I was just reading through the code).

scoder:
I also read it this way, that it's exclusive. It either matches a specific class or the fallback class. While it wouldn't necessarily have to be this way, one advantage is that it makes it possible to exclude certain tags from inheriting the default mixin. If the default was always applied, that would be difficult.

xmo-odoo:
> I also read it this way, that it's exclusive. It either matches a specific class or the fallback class. While it wouldn't necessarily have to be this way, one advantage is that it makes it possible to exclude certain tags from inheriting the default mixin. If the default was always applied, that would be difficult.

That interpretation could make some sense but it seems somewhat inconsistent as it seems to behave differently if the class being augmented is passed through `classes` or is one of the defaults, in that case both mixins will apply.

On the other hand the mixins and classes features don't seem used by the module itself when instantiating its parsers so if the object is not considered part of the public API (is just an internal feature) an other option would be to deprecate those things and leave `HtmlElementClassLookup` as just a registry of "specialised" classes (for form elements)?

Also coming back to

> It's probably also missing tests for comments, PIs and entities, that's why it didn't show so far. Could you please add some?

Sure but there doesn't seem to be much to test aside from testing that Python's inheritance works the way it's specced?

> Thanks. That looks correct. Can't say why it wasn't done like this at the time.

Maybe time or couldn't be arsed, that happens.

xmo-odoo:
> > Sure but there doesn't seem to be much to test aside from testing that Python's inheritance works the way it's specced?
> 
> Or rather that Comment, PI and Entity classes are correctly inserted and still behave as expected when found in the tree, and that XML attribute setting on elements does what it should (there might be tests for that, at least). I mean, the mere fact that your changes broke no tests shows that there were tests missing.

It might also be that the change doesn't break anything because it manipulates the MRO in ways which were not intentionally leveraged (and the old behaviour had to be worked around hence the attributes being re-set). But I see your point.

> I actually wonder if the `set()` method is still needed. It seems redundant now. Maybe it's just there to change the docstring? Might be, then we could still keep it.

It is still useful for the docstring, as normally `set` doesn't accept a `None`, and also to add a default value so on an `HtmlElement` one can call e.g. `e.set('disabled')` and that's equivalent to `<e disabled>`, which is not supported by the XML API. Though I should probably check if there are tests because it looks like this behaviour is more magical than I'd expected, it's not just that (somehow) the element-creation accepts and casts `None` to empty strings for html elements.

scoder:
`Element.set(name, None)` is actually allowed for HTML documents, whether you use `lxml.html` or just the HTML parser in `lxml.etree`. I added that to the docstring in `lxml.etree`. So it's only the default argument value that `lxml.html` adds, which seems helpful for HTML but not for `lxml.etree`.

xmo-odoo:
> `Element.set(name, None)` is actually allowed for HTML documents, whether you use lxml.html or just the HTML parser in lxml.etree.

Right, but without the override that's not visible through the API doc, because you get the information from `_Element`/`ElementBase`, which doesn't really know because it's a special-case based on the parser/document associated with the specific element (if I understand correctly).

xmo-odoo:
@scoder I've one more question: I'm trying to write a test to ensure

> Comment, PI and Entity classes are correctly inserted and still behave as expected when found in the tree

however there seems to be a snag: `resolve_entities` does not exist on the `HTMLParser` and apparently it's an opt-out flag in libxml2, so I can't find a way to actually get an `HtmlEntity` out of a parse.

Also more of a propriety / cleanliness question: `remove_pis` and `remove_flags` are `False` by default, which is what I want, should I assume they remain that or should I use an explicit `HTMLParser` with those flags explicitely set to false, to ignore / discard changes to the default configuration?

xmo-odoo:
Added a pair of tests to check for mixin features (the only one which seems really applicable is `base_url` so I checked that), as well as a few tests around the behaviour of `HtmlElement.set` as I couldn't find any invocation entirely leaving out the value (I also added an explicit `None` and an empty string for good measure).

Also updated the tox file to specifically allow `make`, otherwise some of the envs complain. It still fails locally on 3.8 and 3.10, possibly because everything's installed using pyenv (though that makes it unclear why it'd succeed for 3.5, 3.6, 3.7, and 3.9 which are also installed using pyenv).

src/lxml/html/tests/test_basic.py
```diff
@@ -1,11 +1,51 @@
...
+        self.assertEqual(len(doc), 3)
+        self.assertIsInstance(doc[0], html.HtmlComment)
+        self.assertIsInstance(doc[1], html.HtmlProcessingInstruction)
+        self.assertIsInstance(doc[2], html.HtmlElement)
+        for child in doc:
+            # base_url makes sense on all nodes (kinda) whereas `classes` or
+            # `get_rel_links` not really
+            self.assertEqual(child.base_url, base_url)
+
+    def test_set_extension(self):
```
scoder, line 27: Not sure what you mean with "extension" here.
```suggestion
    def test_set_empty_attribute(self):
```

xmo-odoo: > Not sure what you mean with "extension" here.

I was thinking about the feature being added by / through the HtmlMixin (compared to the regular Element which is what I usually interact with when using lxml) but it's true that since the tests are already in the html context it doesn't make much sense.

Heartcore11 (approved):
src/lxml/html/__init__.py

Heartcore11:
> src/lxml/html/**init**.py
src/lxml/html/__init__.py
src/lxml/html/__init__.py

Heartcore11:
src/lxml/html/__init__.py

Heartcore11:
src/lxml/html/__init__.py

Heartcore11 (commented):
src/lxml/html/__init__.py

Heartcore11 (approved):
src/lxml/html/__init__.py

Heartcore11 (approved):

Heartcore11:
src/lxml/html/__init__.py

src/lxml/html/tests/test_basic.py
Heartcore11: src/lxml/html/__init__.py

Heartcore11 (approved):
https://github.com/lxml/lxml/blob/master/.travis.yml

src/lxml/html/tests/test_basic.py
Heartcore11: > src/lxml/html/__init__.py

https://github.com/lxml/lxml/blob/master/.travis.yml

Heartcore11: > Not sure what you mean with "extension" here.
> ``https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975
>     def test_set_empty_attribute(self):
> ```

Heartcore11: > Not sure what you mean with "extension" here.
> ```suggestion
>     def test_set_empty_attribute(self):
> ```

Heartcore11: > > src/lxml/html/__init__.py
> https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975
>https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975 https://github.com/lxml/lxml/blob/master/.travis.yml

Heartcore11: > Not sure what you mean with "extension" here.
> ```suggestion
>     def test_set_empty_attribute(self):
> ```

https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975

Heartcore11: > > src/lxml/html/__init__.py
> https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975
> https://github.com/lxml/lxml/blob/master/.travis.yml

Heartcore11:
> As the old comment / FIXME from 8132c755adad4a75ba855d985dd257493bccc7fd notes, the mixin should come first for the inheritance to be correct (the left-most class is the first in the MRO, at least if no diamond inheritance is involved).
> 
> Also fix the odd `super` in `HtmlMixin` likely stemming from the incorrect MRO.
> 
> Fixes the inheritance order of all `HTML*` base classes though it probably doesn't matter for other than `HtmlElement`.

https://github.com/lxml/lxml/pull/340

Heartcore11:
> As the old comment / FIXME from 8132c755adad4a75ba855d985dd257493bccc7fd notes, the mixin should come first for the inheritance to be correct (the left-most class is the first in the MRO, at least if no diamond inheritance is involved).
> 
> Also fix the odd `super` in `HtmlMixin` likely stemming from the incorrect MRO.
> 
> Fixes the inheritance order of all `HTML*` base classes though it probably doesn't matter for other than `HtmlElement`.

https://github.com/lxml/lxml/pull/340

Heartcore11:
> As the old comment / FIXME from 8132c755adad4a75ba855d985dd257493bccc7fd notes, the mixin should come first for the inheritance to be correct (the left-most class is the first in the MRO, at least if no diamond inheritance is involved).
> 
> Also fix the odd `super` in `HtmlMixin` likely stemming from the incorrect MRO.
> 
> Fixes the inheritance order of all `HTML*` base classes though it probably doesn't matter for other than `HtmlElement`.

8132c755adad4a75ba855d985dd257493bccc7fd

Heartcore11:
> As the old comment / FIXME from 8132c755adad4a75ba855d985dd257493bccc7fd notes, the mixin should come first for the inheritance to be correct (the left-most class is the first in the MRO, at least if no diamond inheritance is involved).
> 
> Also fix the odd `super` in `HtmlMixin` likely stemming from the incorrect MRO.
> 
> Fixes the inheritance order of all `HTML*` base classes though it probably doesn't matter for other than `HtmlElement`.

src/lxml/html/tests/test_basic.py
Heartcore11: > Not sure what you mean with "extension" here.
> ```suggestion
>     def test_set_empty_attribute(self):
> ```

8132c755adad4a75ba855d985dd257493bccc7fd

Heartcore11: > > src/lxml/html/__init__.py
> 8132c755adad4a75ba855d985dd257493bccc7fd
> https://github.com/lxml/lxml/blob/master/.travis.yml

Heartcore11: > > Not sure what you mean with "extension" here.
> 
> I was thinking about the feature being added by / through the HtmlMixin (compared to the regular Element which is what I usually interact with when using lxml) but it's true that since the tests are already in the html context it doesn't make much sense.

Heartcore11: > Not sure what you mean with "extension" here.
> ```suggestion
>     def test_set_empty_attribute(self):
> ```

Heartcore11: > > Not sure what you mean with "extension" here.
> > ``https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975
> >     def test_set_empty_attribute(self):
> > ```
> 
> 

https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975

Heartcore11: > Not sure what you mean with "extension" here.
> ```suggestion
>     def test_set_empty_attribute(self):
> ```
https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975

Heartcore11:
https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975

Heartcore11:
> > As the old comment / FIXME from 8132c755adad4a75ba855d985dd257493bccc7fd notes, the mixin should come first for the inheritance to be correct (the left-most class is the first in the MRO, at least if no diamond inheritance is involved).
> > 
> > Also fix the odd `super` in `HtmlMixin` likely stemming from the incorrect MRO.
> > 
> > Fixes the inheritance order of all `HTML*` base classes though it probably doesn't matter for other than `HtmlElement`.
> https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975
> 
https://github.com/lxml/lxml/pull/340#pullrequestreview-904505975

PR #341, opened 2022-03-11
Add warning about the global `set_element_class_lookup`

xmo-odoo:
It's a bit of an attractive nuisance, as it basically turns off parser-specific class lookups, which can be surprising to people who have not looked up the mechanism in details (it certainly surprised me).

Also set `inherited-members` in the autodoc config: it's a bit dodgy but because `_BaseParser` is `@cpython.internal` it looks like autodoc can't see it at all, and so anything defined on `_BaseParser` is invisible in the new rst-based doc.

Could not test the docs because I'm hitting https://github.com/cython/cython/issues/3876, does one of the statuses provide a doc to look at?

### DIFF
--- a/src/lxml/classlookup.pxi
+++ b/src/lxml/classlookup.pxi
@@ -549,7 +549,11 @@ cdef void _setElementClassLookupFunction(
 def set_element_class_lookup(ElementClassLookup lookup = None):
     u"""set_element_class_lookup(lookup = None)
 
-    Set the global default element class lookup method.
+    Set the global element class lookup method.
+
+    This defines the main entry point for looking up element implementations.
+    The standard implementation uses the :class:`ParserBasedElementClassLookup`
+    to delegate to different lookup schemes for each parser. 
 
     .. warning::
 
--- a/src/lxml/classlookup.pxi
+++ b/src/lxml/classlookup.pxi
@@ -553,13 +553,16 @@ def set_element_class_lookup(ElementClassLookup lookup = None):
 
     .. warning::
 
-        Setting the default element class to something other than a
-        :class:`ParserBasedElementClassLookup` will prevent
-        :meth:`~lxml.etree.XMLParser.set_element_class_lookup` from working.
+        This should only be changed by applications, not by library packages.
+        In most cases, parser specific lookups should be preferred,
+        which can be configured via
+        :meth:`~lxml.etree.XMLParser.set_element_class_lookup`
+        (and the same for HTML parsers).
 
-        Amongst other issues, this will break :mod:`lxml.html` and
-        :mod:`lxml.objectify`'s extensions as they rely on lookups configured
-        on their respective parsers.
+        Globally replacing the element class lookup by something other than a
+        :class:`ParserBasedElementClassLookup` will prevent parser specific lookup
+        schemes from working. Several tools rely on parser specific lookups,
+        including :mod:`lxml.html` and :mod:`lxml.objectify`.
     """
     if lookup is None or lookup._lookup_function is NULL:
         _setElementClassLookupFunction(NULL, None)
--- a/src/lxml/classlookup.pxi
+++ b/src/lxml/classlookup.pxi
@@ -550,6 +550,16 @@ def set_element_class_lookup(ElementClassLookup lookup = None):
     u"""set_element_class_lookup(lookup = None)
 
     Set the global default element class lookup method.
+
+    .. warning::
+
+        Setting the default element class to something other than a
+        :class:`ParserBasedElementClassLookup` will prevent
+        :meth:`~lxml.etree.XMLParser.set_element_class_lookup` from working.
+
+        Amongst other issues, this will break :mod:`lxml.html` and
+        :mod:`lxml.objectify`'s extensions as they rely on lookups configured
+        on their respective parsers.
     """
     if lookup is None or lookup._lookup_function is NULL:
         _setElementClassLookupFunction(NULL, None)
--- a/doc/api/conf.py
+++ b/doc/api/conf.py
@@ -46,6 +46,7 @@ html_theme_options = {
 autodoc_default_options = {
     'ignore-module-all': True,
     'private-members': True,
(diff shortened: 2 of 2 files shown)


xmo-odoo:
Also the new doc is... a bit of a mess. Would it be OK to submit PRs to commit the apidoc-generated stuff and start hand-crafting the documents in order to make them more logical and stuff? e.g. the `lxml.etree` module starts with 6 pages of exceptions which doesn't make for the most nicest of reads.

scoder:
> Could not test the docs because I'm hitting [cython/cython#3876](https://github.com/cython/cython/issues/3876),

That's long solved, though. That was a problem in 0.29.22, even the legacy stable release of Cython is at 0.29.28 now.

src/lxml/classlookup.pxi
```diff
@@ -550,6 +550,16 @@ def set_element_class_lookup(ElementClassLookup lookup = None):
...
+
+    .. warning::
+
+        Setting the default element class to something other than a
+        :class:`ParserBasedElementClassLookup` will prevent
+        :meth:`~lxml.etree.XMLParser.set_element_class_lookup` from working.
+
+        Amongst other issues, this will break :mod:`lxml.html` and
+        :mod:`lxml.objectify`'s extensions as they rely on lookups configured
+        on their respective parsers.
```
scoder, line 562:
```suggestion
    .. warning::

        This should only be changed by applications, not by library packages.
        In most cases, parser specific lookups should be preferred,
        which can be configured via
        :meth:`~lxml.etree.XMLParser.set_element_class_lookup`
        (and the same for HTML parsers).

        Globally replacing the element class lookup by something other than a
        :class:`ParserBasedElementClassLookup` will prevent parser specific lookup
        schemes from working. Several tools rely on parser specific lookups,
        including :mod:`lxml.html` and :mod:`lxml.objectify`.
```

scoder:
> Also the new doc is... a bit of a mess. Would it be OK to submit PRs to commit the apidoc-generated stuff and start hand-crafting the documents in order to make them more logical and stuff? e.g. the `lxml.etree` module starts with 6 pages of exceptions which doesn't make for the most nicest of reads.

Don't know if there is a way in `apidoc` to change the ordering, but there probably is. I agree that exceptions certainly aren't the most interesting part. Alphabetic ordering is helpful, though, so maybe the normal classes should just go first, followed by functions, followed by exceptions.

Also `lxml.etree`, `lxml.html` and `lxml.objectify` should rather be shown at the top of the list, in that order.

xmo-odoo:
> > Also the new doc is... a bit of a mess. Would it be OK to submit PRs to commit the apidoc-generated stuff and start hand-crafting the documents in order to make them more logical and stuff? e.g. the `lxml.etree` module starts with 6 pages of exceptions which doesn't make for the most nicest of reads.
> 
> Don't know if there is a way in `apidoc` to change the ordering, but there probably is. I agree that exceptions certainly aren't the most interesting part.

Sadly I don't think there's a way for `apidoc` to do that, hence the query about *committing* the files it generates, instead of running apidoc on doc build it'd just be used once to generate the "initial" state (as a scaffolding tool), then those would be edited directly.

The code being committed can still use autodoc though.

src/lxml/classlookup.pxi
xmo-odoo: My only nitpick with this is even for applications it can be be a risky bet or a dead end and should be considered very carefully.

Though of course the application author / user will likely quickly realise they've broken`lxml.html` or `lxml.objectify` if they were using those (obviously), the inter-relation can be quite non-obvious.

A specific admonitions that libraries should never ever touch this is definitely a good idea tho.

xmo-odoo:
Thinking about it further, maybe it would (also?) be a good idea to clarify that the global element lookup is really the first step for resolution? And remove the word "default" from the docstring of `lxml.etree.set_element_class_lookup`?

That kinda gives the impression that *it* is the fallback, but it's the other way around right? When an element proxy needs to be created, lxml invokes `_elementFactory`, which calls `LOOKUP_ELEMENT_CLASS`, which is the object set by `set_element_class_lookup. So that is the first step, and (optionally) calling the parser's lookup' (or not) is *its* responsibility.

scoder:
I'd rather not commit generated files. They just bloat the history, complicate merges and are annoying to work with in various ways. The way to control API documentation in Sphinx seems to be module files and/or TOC files.

There are also [various hacks](https://stackoverflow.com/questions/29385564/customize-templates-for-sphinx-apidoc) that people have come up with.

I'm not particularly tied to sphinx-apidoc. If there's a different solution, I'm open.

src/lxml/classlookup.pxi
```diff
@@ -550,6 +550,19 @@ def set_element_class_lookup(ElementClassLookup lookup = None):
     u"""set_element_class_lookup(lookup = None)
 
     Set the global default element class lookup method.
```
scoder, line 552:
```suggestion
    Set the global element class lookup method.

    This defines the main entry point for looking up element implementations.
    The standard implementation uses the :class:`ParserBasedElementClassLookup`
    to delegate to different lookup schemes for each parser. 
```

scoder:
Thanks

xmo-odoo:
Haha nah thank *you*, the final version is quite a bit better than the original proposal. Cheers.

xmo-odoo:
> I'd rather not commit generated files. They just bloat the history, complicate merges and are annoying to work with in various ways. The way to control API documentation in Sphinx seems to be module files and/or TOC files.
> 
> There are also [various hacks](https://stackoverflow.com/questions/29385564/customize-templates-for-sphinx-apidoc) that people have come up with.
> 
> I'm not particularly tied to sphinx-apidoc. If there's a different solution, I'm open.

Committing generated files is an issue when regularly re-exporting isn't it? Here the output of apidoc would provide a skeleton / starting point, then it wouldn't be used anymore, so there's no more complication than if the initial state had been written by hand. That's how I've used it in the past anyway. Pretty much what the second answer suggests in your link.

PR #342, opened 2022-04-07
Include aarch64 wheel for musllinux SOABI

riconnon:
In order to allow consumers to get a consistent experience in
multi-architecture installs build aarch64 wheel for musllinux SOABI as
well as manylinux


scoder:
Thanks

PR #343, opened 2022-04-15
Allow cross-compiling for Windows ARM64

zooba:
When cross-compiling for Windows ARM64 (the most likely scenario), the `VSCMD_ARG_TGT_ARCH` variable is set to `arm64`. In this case, we still want to grab the ARM64 binaries instead of those matching the host (build) platform.

This variable is the same one that is used inside distutils/setuptools to tell when cross-compiling, so everything else works fine. It is set by default in a Visual Studio cross-compilation environment, but can also be set manually and setuptools will pick up the right settings by itself.

### DIFF
--- a/buildlibxml.py
+++ b/buildlibxml.py
@@ -38,7 +38,7 @@ def download_and_extract_windows_binaries(destdir):
         if release_path in filename
     ]
 
-    if platform.machine() == 'ARM64':
+    if platform.machine() == 'ARM64' or os.getenv('VSCMD_ARG_TGT_ARCH') == 'arm64':
         arch = "win-arm64"
     elif sys.maxsize > 2**32:
         arch = "win64"
--- a/buildlibxml.py
+++ b/buildlibxml.py
@@ -38,6 +38,8 @@ def download_and_extract_windows_binaries(destdir):
         if release_path in filename
     ]
 
+    # Check for native ARM64 build or the environment variable that is set by
+    # Visual Studio for cross-compilation (same variable as setuptools uses)
     if platform.machine() == 'ARM64' or os.getenv('VSCMD_ARG_TGT_ARCH') == 'arm64':
         arch = "win-arm64"
     elif sys.maxsize > 2**32:
--- a/setupinfo.py
+++ b/setupinfo.py
@@ -5,7 +5,7 @@ import os.path
 import subprocess
 from distutils.core import Extension
 from distutils.errors import CompileError, DistutilsOptionError
-from distutils.command.build_ext import build_ext as _build_ext
+from setuptools.command.build_ext import build_ext as _build_ext
 from versioninfo import get_base_dir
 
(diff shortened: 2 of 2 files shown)


buildlibxml.py
```diff
@@ -38,7 +38,7 @@ def download_and_extract_windows_binaries(destdir):
         if release_path in filename
     ]
 
-    if platform.machine() == 'ARM64':
+    if platform.machine() == 'ARM64' or os.getenv('VSCMD_ARG_TGT_ARCH') == 'arm64':
```
scoder, line 41: This thing seems worth a comment to avoid one web search per code read.

setupinfo.py
```diff
@@ -5,7 +5,7 @@
 import subprocess
 from distutils.core import Extension
 from distutils.errors import CompileError, DistutilsOptionError
-from distutils.command.build_ext import build_ext as _build_ext
+from setuptools.command.build_ext import build_ext as _build_ext
```
scoder, line 8: Probably ok to change this since we're importing `setuptools` anyway. What was the reason why this was necessary?

zooba: setuptools has an additional override in its `build_ext.get_ext_filename` to set the right suffix when compiling (in this case, `.cp310-win_arm64.pyd`) that distutils does not. Under normal circumstances, the presence of setuptools is enough to get the right command, so it's transparent, but when you explicitly import it from distutils (even with the setuptools bundled version of distutils) and subclass, then you override it.

Eventually setuptools will merge the two, and I think they consider imports directly from distutils to be deprecated (at least, I hope they do), so those will eventually stop working. But for the transition, they're keeping them working.

buildlibxml.py
zooba:
```suggestion
    # Check for native ARM64 build or the environment variable that is set by
    # Visual Studio for cross-compilation (same variable as setuptools uses)
    if platform.machine() == 'ARM64' or os.getenv('VSCMD_ARG_TGT_ARCH') == 'arm64':
```

setupinfo.py
scoder: Ok. I think we should then import setuptools before `from distutils.core import Extension`, to mark the dependency. Otherwise, this would stop working if the import order in `setup.py` changed.

setupinfo.py
```diff
@@ -5,7 +5,7 @@
 import subprocess
 from distutils.core import Extension
```
scoder, line 6:
```suggestion

from setuptools.command.build_ext import build_ext as _build_ext
from distutils.core import Extension
```

setupinfo.py
scoder, line 8:
```suggestion
```

zooba:
LGTM 👍 Thanks for keeping on this one

scoder:
Thanks

PR #344, opened 2022-05-19
Support PEP 623 (Python 3.12)

methane:
Python 3.12 will remove PyUnicode_AS_DATA and PyUnicode_GET_DATA_SIZE.

### DIFF
--- a/src/lxml/includes/etree_defs.h
+++ b/src/lxml/includes/etree_defs.h
@@ -127,6 +127,19 @@ static PyObject* PyBytes_FromFormat(const char* format, ...) {
 #  define _lx_PySlice_GetIndicesEx(o, l, b, e, s, sl) PySlice_GetIndicesEx(((PySliceObject*)o), l, b, e, s, sl)
 #endif
 
+static const char* unicode_data_and_size(PyObject *unicode, Py_ssize_t *len) {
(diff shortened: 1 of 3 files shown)


scoder:
Thanks, but this should already be resolved by https://github.com/lxml/lxml/commit/bd605086aa053beb35d1bc4e7d3d07f51b93c8e2

methane:
Wow. It is much simpler than mine.

src/lxml/includes/etree_defs.h
```diff
@@ -127,6 +127,19 @@ static PyObject* PyBytes_FromFormat(const char* format, ...) {
...
+    *len = PyUnicode_GET_DATA_SIZE(unicode);
+    return PyUnicode_AS_DATA(unicode);
+#else
+    if (PyUnicode_READY(unicode) < 0) {
+        return NULL;
+    }
+    *len = PyUnicode_KIND(unicode) * PyUnicode_GET_LENGTH(unicode);
+    return (const char*)PyUnicode_DATA(unicode);
+#endif
+}
```
scoder, line 141: I personally prefer not adding a helper function for this now that will be removed again when we drop support for Py3.11, thus refactoring the related code twice. It seems simpler to remove the old API calls at that point.

PR #347, opened 2022-07-14
Correct condition for building docs in CI

janssenhenning:
Previously the `Build Docs` step was always skipped.

I noticed this, when trying to find out, why the link [https://lxml.de/apidoc/objects.inv](https://lxml.de/apidoc/objects.inv) stopped working. This file is used by the `intersphinx` extension to Sphinx to reference objects in other documentations and should be generated by `autodoc`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,7 @@ jobs:
         run: bash ./tools/ci-run.sh
 
       - name: Build docs
-        if: contains( env.EXTRA_DEPS, 'sphinx')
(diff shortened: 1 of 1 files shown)


scoder:
Thanks

scoder:
However, the fix for the problem that you describe is probably this: https://github.com/lxml/lxml/commit/56de38e2745f01427c9cd8c518ba4f50731f388c

scoder:
I've updated the website to include the missing files again.

janssenhenning:
@scoder Thanks 👍

PR #349, opened 2022-08-30
Validate that host_whitelist is not a string

timmc:
Attacker can use `https:///evil.com` to make a malformed "hostless" URL
that would have a `netloc == ''` -- which is `in` any string. Strings
are not documented to be allowed in this config variable anyhow, so just
raise a type error if someone passes in a string by accident.

(This is a breaking change for people who didn't follow the documented
types, but shouldn't affect anyone else.)

New test fails on current master.


src/lxml/html/clean.py
```diff
@@ -242,6 +243,9 @@ def __init__(self, **kw):
                                  "allow_tags and remove_unknown_tags")
             self.remove_unknown_tags = False
 
+        if not isinstance(self.host_whitelist, collection_types):
+            raise TypeError("host_whitelist must be a collection type")
```
scoder, line 247: I think it would be nice to always turn the value into a frozenset (if non-empty).

src/lxml/html/clean.py
```diff
@@ -221,15 +221,16 @@ class Cleaner(object):
     safe_attrs_only = True
     safe_attrs = defs.safe_attrs
     add_nofollow = False
-    host_whitelist = ()
+    host_whitelist = []
```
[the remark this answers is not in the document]
timmc: It's not necessary to change, but I saw a _lot_ of code out there passing in an empty tuple. Passing an empty list would make it less likely that someone accidentally passes in `("example.com")` and is confused about the error. But the guard later on should catch it anyhow!

src/lxml/html/clean.py
timmc: As defensive copying, and for speed of lookups?

scoder: Both, yes. We should generally own config collections that the user passed into the constructor, and not let future modifications leak in. I'm aware that that wasn't done before.

timmc: Hmm, just noticed that you said non-empty -- why so?

timmc: What do you think -- any further changes needed?

scoder:
>Hmm, just noticed that you said non-empty -- why so?

Because it wasn't necessary until I understood that copying the input container would generally be a good idea. Now it's generally necessary. ;-)

src/lxml/html/clean.py
```diff
@@ -206,11 +206,18 @@ def __init__(self, **kw):
         not_an_attribute = object()
         for name, value in kw.items():
             default = getattr(self, name, not_an_attribute)
-            if (default is not None and default is not True and default is not False
-                    and not isinstance(default, (frozenset, set, tuple, list))):
+            if default is None or default is True or default is False:
+                pass
+            elif isinstance(default, (frozenset, set, tuple, list)):
+                # Catch common error of passing ('host') instead of a tuple.
+                if isinstance(default, str):
```
scoder, line 213:
```suggestion
                if isinstance(value, str):
```

src/lxml/html/tests/test_clean.py
```diff
@@ -271,6 +271,19 @@ def test_formaction_attribute_in_button_input(self):
...
+        # creates a host_whitelist of the empty string; a malformed triple-slash
+        # URL has an "empty host" according to urlsplit, and `"" in ""` passes.
+        # So, don't allow user to accidentally pass a string for host_whitelist.
+        html = '<div><iframe src="https:///evil.com/page"></div>'
+        with self.assertRaises(TypeError) as exc:
+            # If this were the intended `("example.com",)` the expected
+            # output would be '<div></div>'
+            Cleaner(frames=False, host_whitelist=("example.com")).clean_html(html)
+
+        self.assertEqual(("host_whitelist must be a collection type",), exc.exception.args)
```
scoder, line 285:
```suggestion
        self.assertEqual(("Expected a collection, got str: host_whitelist='example.com'",), exc.exception.args)
```

scoder:
Thanks

PR #350, opened 2022-09-12
Try and preserve the structure of the html during a diff

lonetwin:
There exists a bug in the current `htmldiff` code, where by the generated diff
changes the structure of the html:

```python
>>> from lxml.html import diff
>>> a = "<div id='first'>some old text</div><div id='last'>more old text</div>"
>>> b = "<div id='first'>some old text</div><div id='middle'>and new text</div><div id='last'>more old text</div>"
>>> diff.htmldiff(a, b)
('<div id="middle"> <div id="first"><ins>some old text</ins></div><ins>and new</ins> <del>some old</del> text</div><div id="last">more old text</div>')
>>>

```

This patchset is an attempt to fix that issue.

### DIFF
--- a/src/lxml/html/diff.py
+++ b/src/lxml/html/diff.py
@@ -234,22 +234,38 @@ def expand_tokens(tokens, equal=False):
 def merge_insert(ins_chunks, doc):
     """ doc is the already-handled document (as a list of text chunks);
     here we add <ins>ins_chunks</ins> to the end of that.  """
-    # Though we don't throw away unbalanced_start or unbalanced_end
-    # (we assume there is accompanying markup later or earlier in the
-    # document), we only put <ins> around the balanced portion.
+
+    ins_chunks = list(ins_chunks)
     unbalanced_start, balanced, unbalanced_end = split_unbalanced(ins_chunks)
-    doc.extend(unbalanced_start)
-    if doc and not doc[-1].endswith(' '):
-        # Fix up the case where the word before the insert didn't end with 
-        # a space
-        doc[-1] += ' '
-    doc.append('<ins>')
-    if balanced and balanced[-1].endswith(' '):
-        # We move space outside of </ins>
-        balanced[-1] = balanced[-1][:-1]
-    doc.extend(balanced)
-    doc.append('</ins> ')
+
+    # add ins tag to the doc
+    leading_space = "" if doc and doc[-1][-1] == " " else " "
+    doc.append("{0}<ins>".format(leading_space))
+
+    for chunk in ins_chunks:
+        if chunk in unbalanced_start:
+            leading_space = "" if chunk[-1] == " " else " "
+            if doc and '<ins>' in doc[-1]:
+                doc.pop()
+                doc.extend([
+                    chunk,
+                    "{0}<ins>".format(leading_space)
+                ])
+            else:
+                trailing_space = "" if chunk[0] == " " else " "
+                doc.extend([
+                    "</ins>{0}".format(trailing_space),
+                    chunk,
+                    "{0}<ins>".format(leading_space)])
+        elif chunk not in unbalanced_end:
+            doc.append(chunk)
+
+    if doc and doc[-1] and doc[-1][-1] == " ":
+        doc[-1] = doc[-1][:-1]
+
+    doc.append("</ins> ")
     doc.extend(unbalanced_end)
+
 
 # These are sentinels to represent the start and end of a <del>
 # segment, until we do the cleanup phase to turn them into proper
--- a/src/lxml/html/tests/test_diff.txt
+++ b/src/lxml/html/tests/test_diff.txt
@@ -87,6 +87,13 @@ Whitespace is generally ignored for the diff but preserved during the diff:
     second
      <ins>third</ins> </pre>
 
+Ensure we don't preserve the structure of the html during the diff:
+
+    >>> a = "<div id='first'>some old text</div><div id='last'>more old text</div>"
+    >>> b = "<div id='first'>some old text</div><div id='middle'>and new text</div><div id='last'>more old text</div>"
+    >>> print(htmldiff(a, b))
+    <div id="first"><ins>some old text</ins></div> <div id="middle"> <ins>and new</ins> <del>some old</del> text</div><div id="last">more old text</div>
+
 The sixteen combinations::
 
 First "insert start" (del start/middle/end/none):


src/lxml/html/tests/test_diff.txt
```diff
@@ -54,8 +54,8 @@ As a special case, changing the href of a link is displayed, and
 images are treated like words:
 
     >>> pdiff('<a href="http://yahoo.com">search</a>', '<a href="http://google.com">search</a>')
-    <a href="http://google.com">search <ins> Link: http://google.com</ins>
-    <del> Link: http://yahoo.com</del> </a>
+    <a href="http://google.com">search <ins> Link: http://google.com
+    </ins> <del> Link: http://yahoo.com</del> </a>
```
scoder, line 58: Was this a necessary change? It doesn't feel right.

src/lxml/html/tests/test_diff.txt
```diff
@@ -87,6 +87,13 @@ Whitespace is generally ignored for the diff but preserved during the diff:
     second
      <ins>third</ins> </pre>
 
+Ensure we don't preserve the structure of the html during the diff:
```
scoder, line 90: "don't" ?

src/lxml/html/diff.py
```diff
@@ -234,22 +234,38 @@ def expand_tokens(tokens, equal=False):
...
-    if balanced and balanced[-1].endswith(' '):
-        # We move space outside of </ins>
-        balanced[-1] = balanced[-1][:-1]
-    doc.extend(balanced)
-    doc.append('</ins> ')
-    doc.extend(unbalanced_end)
+
+    for chunk in ins_chunks:
+        if chunk in balanced:
+            doc.append(chunk)
```
scoder, line 250: This seems risky. I'm not sure that we can assume chunks to be unique enough to just do an "in" test here.

Maybe it could help to put the three parts back together, but give each chunk a marker like `(chunk, "balanced")` and `(chunk, "unbalanced")`, and then do something based on the marker. Or look at the index boundaries where we switch between the unbalanced and balanced parts.

src/lxml/html/tests/test_diff.txt
lonetwin: Ah sorry, that was a brain frat ! Good catch. I think I might have wanted to add something like 'Ensure we don't mess with the structure ...' and instead decided to go with the opposite phrasing 'Ensue we preserve ...' but obviously messed up. I'll fix in the next commit.

src/lxml/html/tests/test_diff.txt
lonetwin: tbh, didn't feel right to me either but this test breaks without it, and I didn't have much luck trying to figure out why. Initial instinct was that it has something to do with the `re.sub` in the `pwrapped` definition at the top of the test file (possibly the `+` in the re over matches ?). Apologies, I didn't try digging deeper but I'll give it a shot now.

src/lxml/html/diff.py
lonetwin: Hmm, agreed that we can't assume the chunks to be unique enough to just do an 'in' test here. I'll think about your suggestion and give this another go. Thanks for your review.

src/lxml/html/tests/test_diff.txt
lonetwin: Done

src/lxml/html/diff.py
lonetwin: We ran into an issue rising out of precisely the problem you caught ! The visible effect being that the `<ins>` tags were not being added at places where they ought to be.

I've updated the approach by changing `split_unbalanced` to *optionally* return a list of tuples containing the `(<index-in-the-input=chunks-list>, <chunk>)`. This then serves to uniquely identify the chunk across the balanced/unbalanced lists ...which then makes the `in` operator reliable.

I've defaulted to retain the existing behavior of `split_unbalanced` since it is also called from `cleanup_delete` and I didn't want to touch that function.

Hopefully, you find the changes suitable.

src/lxml/html/tests/test_diff.txt
lonetwin: So the underlying cause for this appears to be an extra trailing whitespace that appears to get added to the `href_token` which pushes the line length of the string to over 70 (the line length limit of textwrap, used by pdiff). Here are the diffs with the current lxml code and one taken with this patchset:

current lxml diff:

`<a href="http://google.com">search <ins> Link: http://google.com</ins> <del> Link: http://yahoo.com</del> </a>`

diff with this patchset:
`<a href="http://google.com">search <ins> Link: http://google.com </ins> <del> Link: http://yahoo.com</del> </a>`

What I'm not entirely sure about is why the trailing whitespace which ideally *should* also be present in the current lxml diff (according to call to `href_token` in `fixup_chunks`), isn't.

In any case, would you consider this as a regression in behavior worthy of a fix or does this fix of the test case suffice ?

scoder: What strikes me is that there is a space before the closing `</ins>` but not before the closing `</del>`. Given that there is a space after both of the opening `<ins>` and `<del>`, it's debatable whether the new behaviour isn't really better than the old. But I would expect both tag types to be consistent, at least.

src/lxml/html/diff.py
```diff
@@ -338,31 +342,32 @@ def split_unbalanced(chunks):
...
-                end.append(chunk)
+                end.append(chunk_to_add)
         else:
-            tag_stack.append((name, len(balanced), chunk))
+            tag_stack.append((name, len(balanced), chunk_to_add))
             balanced.append(None)
     start.extend(
-        [chunk for name, pos, chunk in tag_stack])
+        [chunk_to_add for name, pos, chunk_to_add in tag_stack])
     balanced = [chunk for chunk in balanced if chunk is not None]
```
scoder, line 371: The condition in this line probably doesn't work any more if `with_idx=True`.

src/lxml/html/tests/test_diff.txt
```diff
@@ -87,12 +87,20 @@ Whitespace is generally ignored for the diff but preserved during the diff:
...
+    >>> pdiff(a, b)
+    <div id="first"><ins>some old text</ins></div> <div id="middle">
+    <ins>and new</ins> <del>some old</del> text</div><div id="last">more
+    old text</div>
+    >>> a = "<div><p>Some text that will change</p><p>Some tags will be added</p></div>"
+    >>> b = "<div><div><p>Some text that has changed a bit</p><p>All of this is new</p></div></div>"
+    >>> pdiff(a, b)
+    <div><div><p>Some text that <ins>has changed a bit</ins> </p>
+    <p><ins>All of this is new</ins></p> </div> <del>will
+    change</del><p><del>Some tags will be added</del></p> </div>
```
scoder, line 103: This diff looks weird. Is it getting confused by the nested `<div>`s ?

src/lxml/html/tests/test_diff.txt
lonetwin: I agree. I'll investigate and attempt to make them consistent.

src/lxml/html/tests/test_diff.txt
lonetwin: Yes, that's correct. In this specific case, the rendered diff looks 'ok' to me (with the insertions preceding the deletions) but yeah, I imagine in larger context it might be less than ideal. I'm going to take another stab at this.

src/lxml/html/tests/test_diff.txt
lonetwin: I've addressed this issue by removing the inclusion of the trailing whitespace during the creation of `href_token` instance which adds the text `Link: ...`. The reasoning for this is that the only time this piece of string would be displayed would be when there's a change, which necessarily means it would be handled in the `merge_insert` function.

By making this change, I no longer needed to update the existing test case.

lonetwin:
@scoder gentle reminder - would you mind reviewing this PR once more ? (related: do you expect that comments are marked 'Resolve conversation' by the submitter if the review comment was addressed ?)

scoder:
I took the liberty to rewrite most of the original algorithm with less list copying and more "stick to the tag order". Took me a while to get right, but I think it's mostly an improvement, overall. It's probably also faster.

https://github.com/lxml/lxml/pull/463

@lonetwin, could you please try that PR on your side?

lonetwin:
> I took the liberty to rewrite most of the original algorithm with less list copying and more "stick to the tag order". Took me a while to get right, but I think it's mostly an improvement, overall. It's probably also faster.
> 
> #463
> 
> @lonetwin, could you please try that PR on your side?

Sure yeah. Will do and I'll report back on that PR.

scoder:
> Sure yeah. Will do and I'll report back on that PR.

@lonetwin, were you able to try it?

PR #352, opened 2022-09-23
Decreasing cyclomatic complexity of code

natalycod:
I used RADON-utilite and found function printSetupTimes with complexity B (6). I get rid of extra for and decrease complexity to A (5)


scoder:
Thanks, but a) this is not an important module, and b) I do not consider the change an improvement. The mere fact that it reduces some kind of automatic measure does not mean that it becomes easier to read. The new version seems substantially harder to understand.

PR #353, opened 2022-10-13
Xpath with namespace and position

knit-bee:
I noticed that it is not possible to use `elem.find` or `elem.findall` with an xpath that contains position indices if the method is called with the `namespaces` argument.
This behavior has also been reported in [Bug #1873886](https://bugs.launchpad.net/lxml/+bug/1873886).

It appears that during the tokenization of the xpath, the numbers are treated as tags, i.e. they are concatenated with the default namespace (during function calls with namespaces). This results in a wrong path imo.
For example:
```py
>>> from lxml import etree
>>> doc = etree.XML("""
      <foo xmlns="http://example.com/foo">
        <bar>baz</bar>
      </foo>""")
>>> path = "./bar[1]"
>>> doc.find(path, namespaces={None:"http://example.com/foo"})
None
```
The target element is not found here because the path that is used is effectively:
`./{http://example.com/foo}bar[{http://example.com/foo}1]`

Changes:
- I added a check during the tokenization of the xpath to determine whether the processed `tag` is a number to avoid concatenation with the namespace.

### DIFF
--- a/src/lxml/_elementpath.py
+++ b/src/lxml/_elementpath.py
@@ -87,7 +87,10 @@ def xpath_tokenizer(pattern, namespaces=None):
                 except KeyError:
                     raise SyntaxError("prefix %r not found in prefix map" % prefix)
             elif default_namespace and not parsing_attribute:
-                yield ttype, "{%s}%s" % (default_namespace, tag)
+                if tag.isnumeric():
+                    yield ttype, tag
+                else:
+                    yield ttype, "{%s}%s" % (default_namespace, tag)
             else:
                 yield token
             parsing_attribute = False
--- a/src/lxml/_elementpath.py
+++ b/src/lxml/_elementpath.py
@@ -86,11 +86,8 @@ def xpath_tokenizer(pattern, namespaces=None):
                     yield ttype, "{%s}%s" % (namespaces[prefix], uri)
                 except KeyError:
                     raise SyntaxError("prefix %r not found in prefix map" % prefix)
-            elif default_namespace and not parsing_attribute:
-                if tag.isnumeric():
-                    yield ttype, tag
-                else:
-                    yield ttype, "{%s}%s" % (default_namespace, tag)
+            elif default_namespace and not parsing_attribute and not tag.isdecimal():
+                yield ttype, "{%s}%s" % (default_namespace, tag)
             else:
                 yield token
             parsing_attribute = False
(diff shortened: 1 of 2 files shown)


src/lxml/tests/test_elementpath.py
```diff
@@ -285,6 +325,233 @@ def test_find(self):
...
+        self.assertEqual(
+            etree.ElementTree(elem).find("tag[1]", namespaces=namespaces).tag, "{nsnone}tag"
+        )
+
+        self.assertEqual(elem.findtext("tag[1]", namespaces=namespaces), "text")
+        self.assertTrue(elem.findtext("tag[2]", namespaces=namespaces) in {None, ''})
+        self.assertEqual(
+            summarize_list(elem.findall("tag", namespaces=namespaces)),
+            ["{nsnone}tag", "{nsnone}tag"],
+        )
```
scoder, line 380: Given how many examples you add here, this seems worth its own custom assert method: `.assertFindallEqual(element, path, expected, namespaces=None)`.

Also, do we actually need to add these tests? ISTM that we could get away with running the existing tests three times, once without namespaces dict, once with an empty one, and once with a non-empty one.

src/lxml/_elementpath.py
```diff
@@ -87,7 +87,10 @@ def xpath_tokenizer(pattern, namespaces=None):
                 except KeyError:
                     raise SyntaxError("prefix %r not found in prefix map" % prefix)
             elif default_namespace and not parsing_attribute:
```
scoder, line 89: I think it's enough to do this:
```suggestion
            elif default_namespace and not parsing_attribute and not tag.isdecimal():
```

scoder (commented):
Thanks. I think we should reduce the amount of code that we add here, though.

knit-bee:
Thank you for the review, I will try to include your suggestions next week.

src/lxml/_elementpath.py
```diff
@@ -87,7 +87,10 @@ def xpath_tokenizer(pattern, namespaces=None):
...
-                yield ttype, "{%s}%s" % (default_namespace, tag)
+                if tag.isnumeric():
+                    yield ttype, tag
+                else:
+                    yield ttype, "{%s}%s" % (default_namespace, tag)
```
scoder, line 93:
```suggestion
                yield ttype, "{%s}%s" % (default_namespace, tag)
```

scoder:
Thanks for this PR, I have integrated it in https://github.com/lxml/lxml/commit/e950468e00ccf4fcdc97e609a8b94e6dacd81481 but implemented the tests differently.

PR #354, opened 2022-10-14
Clarify license

rettinghaus:
This is small update to the license text clarifying that it is indeed version 3 by using the [choosealicense template](https://choosealicense.com/licenses/bsd-3-clause/).


scoder:
Both PyPI and github seem to list the license correctly and clearly, so all I changed was to add a license title in https://github.com/lxml/lxml/commit/96c504f7f4d3d71730261f4cb0711345c19cca33

I'd like to avoid changing something as basic as the name of the license file without reason since you never know who is using the sources in what way.

But thanks for bringing this up.

PR #355, opened 2022-10-20
AppVeyor CI: Update to Visual Studio 2022 image

EwoutH:
This PR updates the AppVeyor configuration to use the Visual Studio 2022 image.

Update 2022-11-07: The Python 3.11 part of this PR has moved to a new PR, #360.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 version: 1.0.{build}
-image: Visual Studio 2019
+image: Visual Studio 2022
 
 environment:
   matrix:


EwoutH:
It looks like the AppVeyor build has been failing for quite some time. I openend a bug report at https://bugs.launchpad.net/lxml/+bug/1993962.

@scoder Do you have any idea what's going on?

EwoutH:
After the fix in 77413c28, it processes further, and it looks like it now fails with this error:

```
fatal error C1047: The object or library file 'libs\iconv-1.14.win64\lib\iconv_a.lib' was created by a different version of the compiler than other objects like 'build\temp.win-amd64-cpython-310\Release\src\lxml\etree.obj'; rebuild all objects and libraries with the same compiler
```

scoder:
[libiconv](https://savannah.gnu.org/projects/libiconv/) is maintained (latest release was last May) and used by libxml2.

EwoutH:
Thanks for l8nking to the right repository. It looks like the latest stable libiconv version is 1.17, and we’re using 1.14. How can we upgrade?

scoder:
> How can we upgrade?

There's a repository with git subrepos for the dependencies. But libiconv remains at 1.14 from where it's forked. So it won't be trivial, I guess. Not sure if we *must* upgrade, though. It might be enough to rebuild the libs in their current state with the right compiler.
https://github.com/lxml/libxml2-win-binaries

EwoutH:
Would you be able to take a look at that?

jchalupka-pan:
I am not an expert in building stuff (for long time got stuck with pure Python), but maybe it will help at least a little bit. I know that it is not a "production solution", but maybe something can be used, since I have built the `lxml` wheel for Windows and Python 3.11 with BuildTools v143 somewhere last week.

1. I used master branch from https://github.com/lxml/libxml2-win-binaries.
2. Applied the patch provided on repo.
3. Find and replace the build tools version in file `libiconv\MSVC16\libiconv_static\libiconv_static.vcxproj` from `v142` to `v143`.
4. Changed the path to `vcvars.bat` in `build.ps1`
5. Build libraries, which can be used while building `lxml` wheel.

Building the `lxml` wheel required a hax to hardcode paths to zips and not use those from https://api.github.com/repos/lxml/libxml2-win-binaries/releases.  I just modified the function `download_and_extract_windows_binaries` from `buildlibxml.py`.

EwoutH:
I split the Python 3.11 part off to a new PR, #360. I would like to request to merge and backport that one first, since it already works on the VS 2019 image.

It would still be nice to get libiconv updated and move to the Visual Studio 2022 image, especially because it has better Arm support. However, that can wait until #360 is merged and Python 3.11 wheels are on PyPI.

cclauss:
Do these tests pass now that #365 has landed?

schlenk:
This may be related to the binary compatibility rules for static libraries and different Visual Studio versions:

https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170#restrictions

> Static libraries or object files compiled using the [/GL (Whole program optimization)](https://learn.microsoft.com/en-us/cpp/build/reference/gl-whole-program-optimization?view=msvc-170) compiler switch or linked using [/LTCG (Link-time code generation)](https://learn.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=msvc-170) aren't binary-compatible across versions, including minor version updates.

So if the prebuilt static libraries of iconv/zlib do not match the compiler used for the final linkage it breaks. You would need to provide static libs for each compiler/toolset version that you want to support.

scoder:
Closing this PR since we're no longer using appveyor.

PR #356, opened 2022-10-20
GHA wheel CI: Update images, used actions and Python version

EwoutH:
A bit of maintenance on the GitHub Actions wheel CI:
 - Update the used Ubuntu and macOS images to the latest versions, and enable the Windows run
 - Update the used actions to their latest versions
 - Use Python 3.10 to build wheels
 - Add Python 3.11 run

Part of [Bug #1977998](https://bugs.launchpad.net/lxml/+bug/1977998).

PR #358, opened 2022-11-02
Add Dependabot configuration for GitHub Actions updates

EwoutH:
Add a Dependabot configuration that checks once a week if the GitHub Actions are still using the latest version. If not, it opens a PR to update them.

It will actually open very few PRs, since we only have major versions specified (like v3), so only on a major v4 release it will update and open a PR.

This will basically automate the majority of PRs like #356.

See [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"


cclauss:
While https://github.com/dependabot/dependabot-core/issues/2804 remains unresolved, I am not a fan of Depenabot.  You can see its ___crazy spam alerts___ on [my fork of CPython](https://github.com/notifications?query=repo%3Acclauss%2Fcpython+)

EwoutH:
That's a valid concern, especially for such a large project like CPython.

What does help for lxml, is that mostly only major versions are specified instead of patch versions (so `@v2` instead of `@v2.3.4`), so only a PR is opened when a major version update is available, which won't be that often.

They're [working on](https://github.com/dependabot/dependabot-core/issues/5914#issuecomment-1282872261) disabling it on forks, so we can also wait a bit to see where that leads to.

> We're actively working on this and a permanent fix should land in the next couple of weeks. We'll share more around the time we''re ready to roll this out

Another option would be using a GitHub Actions workflows with an conditional statement that it only runs on this repository (`lxml/lxml`).

hugovk:
Big +1 for automating this! 

This repo has 515 forks.

I switched all my repos over to [Renovate](https://docs.renovatebot.com/), which doesn't spam forks (by default) and is much more configurable, for example, you can group updates into a single PR and apply labels. 

I could help with that if you'd like. 

Having said that, it's really good to hear GitHub are finally working on a fix, so if you prefer Dependabot I'd definitely hang on a few weeks for that.

scoder:
Thanks. I agree that we should wait a bit more before merging this. I'm annoyed by dependabot's fork-problem, too.

hugovk:
GitHub have partially fixed this.

Dependabot won't be enabled for new forks, but is still enabled for existing forks, which can be individually disabled via their settings:

https://github.com/dependabot/dependabot-core/issues/2804#issuecomment-1306143227

cclauss (approved):

cclauss:
I have now disabled Dependabot on numerous forks and it is easy to do and works as expected.

I think this can move forward if others agree.

EwoutH:
Hi @scoder! Dependabot's spamming forks issue is now fully resolved, I've observed many repositories with it enabled and it hasn't caused any problems. I think it's ready to be merged.

hugovk:
Well, fully resolved for *new* forks.

Old forks still get spammed, but it can be manually turned off with a few clicks, if you know where to look. 

https://github.com/lxml/lxml/pull/358#issuecomment-1308737341

EwoutH:
@hugovk I asked that specifically in that thread (now burried in references), but it only hinders existing forks as they _already have_ a `dependabot.yml` file. Since that isn't the case for forks of this repo (since lxml never had one, this isn't an issue for us, and we're good.

scoder:
Thanks!

hugovk:
@EwoutH Thanks for clearing that up!

PR #359, opened 2022-11-05
Fix wheel building for ARM Mac

fiahil:
This PR partially fix building the lxml wheel on the new Macs.

For some reason I still need to override `ARCHFLAGS="-arch arm64"` to allow it to link it properly.


scoder:
Thanks, although that only helps when users actually build on ARM macs themselves. It doesn't help with building and providing "universal2" ARM/x64 wheels.

EwoutH:
Thanks for this!

One minor nitpick, I think `MACOSX_DEPLOYMENT_TARGET` could be set to a newer value than `"10.6"` for the arm builds. macOS 11.0 Big Sur was the first release to support Apple Silicon, so I think it could be set to `"11.0"` on those builds.

PR #361, opened 2022-11-07
Drop Python 2.7, 3.5 and 3.6, require Python >= 3.7

EwoutH:
Python 2.7, 3.5 and 3.6 are all [end-of-life](https://endoflife.date/python) and don't receive security updates anymore. Dropping those Python versions and raising the minimum required Python version allows to clean up a lot of legacy code and use modern syntax, like [f-strings](https://peps.python.org/pep-0498/).

It also reduced the CI matrix significantly, which reduces CI and build time. Migrating to [cibuildwheel](https://github.com/pypa/cibuildwheel) will also be easier, since they only support Python 3.6+ anyways.

The most important change is bumping the `python_requires` in setup.py, which is what setuptools and PyPI detect. This means that for new releases Python 3.7 or higher is required. Users still on older Pythoin versions can keep using the current releases.

This PR consists of 4 commits, each upgrading a different part of the codebase. Read the commit messages of each commit for details. Please do not squash when merging, since every commit does something different and for debugging purposes it's nice to keep them separate if anything unexpected breaks in the future.


src/lxml/_elementpath.py
```diff
@@ -83,11 +82,11 @@ def xpath_tokenizer(pattern, namespaces=None):
                 try:
                     if not namespaces:
                         raise KeyError
-                    yield ttype, "{%s}%s" % (namespaces[prefix], uri)
+                    yield ttype, f"{{{namespaces[prefix]}}}{uri}"
```
scoder, line 85: I prefer not using f-strings in this specific case. The triple braces make the expression much more complex than the original one. And recent Python (and Cython) versions will convert this into an f-string anyway.

src/lxml/html/diff.py
```diff
@@ -473,7 +467,7 @@ def __new__(cls, text, pre_tags=None, post_tags=None, trailing_whitespace=""):
         return obj
 
     def __repr__(self):
-        return 'token(%s, %r, %r, %r)' % (_unicode.__repr__(self), self.pre_tags,
+        return 'token({}, {!r}, {!r}, {!r})'.format(_unicode.__repr__(self), self.pre_tags,
                                           self.post_tags, self.trailing_whitespace)
```
scoder, line 471: Interesting. Why not an f-string here?

src/lxml/tests/test_dtd.py
```diff
@@ -66,14 +63,12 @@ def test_dtd_parse_file_not_found(self):
 
     def test_dtd_parse_valid(self):
         parser = etree.XMLParser(dtd_validation=True)
-        xml = ('<!DOCTYPE a SYSTEM "%s"><a><b/></a>' %
-               fileUrlInTestDir("test.dtd"))
+        xml = f"<!DOCTYPE a SYSTEM \"{fileUrlInTestDir('test.dtd')}\"><a><b/></a>"
```
scoder, line 66: You can use single-quoted f-strings instead of backslash escapes here.

tools/xpathgrep.py
```diff
@@ -143,7 +142,7 @@ def iter_input(input, filename, parser, line_by_line):
                         yield et.ElementTree(et.fromstring(line, parser))
             else:
                 yield et.parse(input, parser)
-        except IOError:
+        except OSError:
```
scoder, line 145: This looks like a functional change. (Same below.)

update-error-constants.py
```diff
@@ -114,8 +113,8 @@ def main(doc_dir):
         constants.sort(key=operator.itemgetter(1))
         pxi_name, prefix = ENUM_MAP[enum_name]
 
-        append_pxd(ctypedef_indent + 'ctypedef enum %s:' % enum_name)
-        append_pxi('cdef object %s = """\\' % pxi_name)
+        append_pxd(ctypedef_indent + f'ctypedef enum {enum_name}:')
+        append_pxi(f'cdef object {pxi_name} = ""\"\\')
```
scoder, line 117:
```suggestion
        append_pxd(f'{ctypedef_indent}ctypedef enum {enum_name}:')
        append_pxi(f'cdef object {pxi_name} = """\\')
```

versioninfo.py
```diff
@@ -65,16 +65,15 @@ def create_version_h():
...
+        with open(file_path) as version_h:
+            if (f'"{lxml_version}"') in version_h.read(100):
                 return
 
     with open(file_path, 'w') as version_h:
-        version_h.write('''\
-#ifndef LXML_VERSION_STRING
-#define LXML_VERSION_STRING "%s"
+        version_h.write(f'''#ifndef LXML_VERSION_STRING
+#define LXML_VERSION_STRING "{lxml_version}"
```
scoder, line 74:
```suggestion
        version_h.write(f'''\
#ifndef LXML_VERSION_STRING
#define LXML_VERSION_STRING "{lxml_version}"
```

scoder (commented):
First of all, thanks, that's a lot of nice modernisations in one PR and looks like quite some effort.

However, you are touching a lot of files that haven't been touched in long years, and probably won't be touched for a few more. I don't think it's worth modernising their code.

Some of the changes can already be done without dropping legacy support, e.g. using f-strings in Cython files (which, BTW, you didn't touch at all with your changes, although they are the most active and important ones), dict/set comprehensions or 'b' string prefixes.

It seems you broke the encoding in the selftest.py module with your edits.

Overall, I don't really see why we should drop backwards compatibility right now. It doesn't feel like it helps much, but creates a lot of code churn and upsets some users. There are still more than 30K downloads of lxml for Python 2.7 per day, a couple of thousand for 3.5 and six digit download counts for 3.6. I'm usually more a friend of gradual code adaptation along with changes, and of dropping backwards compatibility when it starts hurting more than helping. It doesn't really hurt, currently.

scoder:
Just FYI, these are the PYPI download stats of the last 6 months:
```
2.7 :      28787.7 / day
3.2 :         11.9 / day
3.3 :         23.7 / day
3.4 :        210.3 / day
3.5 :       4928.7 / day
3.6 :     261071.1 / day
3.7 :     600840.8 / day
3.8 :     618036.3 / day
3.9 :     148260.3 / day
3.10:      60504.1 / day
3.11:        818.5 / day
3.12:          7.3 / day
null:     139268.1 / day
```

EwoutH:
Thanks for getting back and reviewing.

I think the big question is, how many of those users on Python 2.7, 3.5 and 3.6 are actively looking for new lxml features, instead of keeping legacy programs running with a current lxml version?

Upgrading the required Python version isn't done for users, mainly for developers and maintainers, because it makes developing new features and maintaining them easier.

> dropping backwards compatibility when it starts hurting more than helping.

One specific area will be wheel building. cibuildwheels doesn't support Python 3.5 and older, so 2.7 and 3.5 can't migrate. 3.6 support is currently still there, but won't be for long. That means keeping (at least) two systems in the air to build wheels. 

If wheel building is dropped on older Python versions, not upgrading the required versions will cause `pip` to install the latest compatible version, for which is no wheel available. This means users that expect a wheel don't get one.

So I think updating the required version before infrastructure starts breaking is a good practice.

EwoutH:
@scoder Thanks for tagging and releasing [4.9.2](https://github.com/lxml/lxml/releases/tag/lxml-4.9.2)! With that released, would you consider dropping Python 2.7 and 3.5 (not 3.6)?

In that case I would like to give it a try to start using [cibuildwheel](https://github.com/pypa/cibuildwheel) for wheelbuilding. That should reduce maintenance effort a lot.

src/lxml/_elementpath.py
EwoutH: I don't this the triple braces are needed here? I could manually adjust that.

src/lxml/html/diff.py
EwoutH: I think it didn't recognize it, I could manually convert it.

tools/xpathgrep.py
EwoutH: This is a modification from [PEP 3151](https://peps.python.org/pep-3151/#confusing-set-of-os-related-exceptions), implemented in Python 3.3. It's only a naming change

PR #364, opened 2022-12-02
GitHub Action: python3.11 -m pip install lxml

cclauss:
Installing lxml on Python 3.11 works on Linux and macOS but not yet on Windows.

This _temporary_ GitHub Action demonstrates how installations fail:
* [ ] On Windows, `pip install lxml` from this repo
* [ ] On Windows, `pip install lxml` [from PyPI](https://pypi.org/project/lxml)

This could run on `workflow_dispatch` instead of on `push` and `pull_request` but then I would not have the rights to run it.

### DIFF
--- a/.github/workflows/python3.11_-m_pip_install_lxml.yml
+++ b/.github/workflows/python3.11_-m_pip_install_lxml.yml
@@ -1,3 +1,7 @@
+# GitHub Action to prove a failure still exists...
+# * lxml still cannot be successfully installed on Python 3.11 on Windows.
+# When this GitHub Action passes then it can be deleted.
+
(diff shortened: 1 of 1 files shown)


EwoutH:
Testing `pip install` (from PyPI) is only useful right after the deploy of a new release. So I would probably make it a dependent workflows after a "Deploy to PyPI" workflow.

So this could be useful, as an extension of the `wheels.yml` workflow, if that would actually deploy wheels to PyPI. But currently that doesn't, the wheels are only uploaded as artifacts to GitHub, and then uploaded manually I believe.

Running on each pull request and push is doesn't do anything useful anyways.

PR #365, opened 2022-12-03
Fix build failed on windows server 2022

MrBenHacker:
Hello,
I was looking since 3.11 release a wheel for my windows OS. Since yesterday's commit, it looked almost working except an issue with Visual Studio 2022. I was actually using it when installing via pip and also break on my local machine.

Tested few workflows on a branch and just realized someone had the same fix already 😄 (https://github.com/lxml/lxml/pull/356#issuecomment-1324905890).

Hopefully a wheel can be published on pypi before waiting for the next lxml release :)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -133,7 +133,7 @@ jobs:
       fail-fast: false
 
       matrix:
-        os: [macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


cclauss:
A passing test on AppVeyor (like #355) or GitHub Actions Windows (like #364) would help a lot.

PR #366, opened 2023-01-30
Make resolve_entities require opt-in

clayg:
Please forgive me if my attempt to be helpful feels ungreatful - I'm probably feeling frustrated or insecure having just read up on  https://bugs.launchpad.net/swift/+bug/1998625 ... and then to realize how long other people have been dealing with this https://bugs.launchpad.net/keystone/+bug/1100282 ... which makes it feel even worse.  Like I *SHOULD* have known.  But I didn't.

I remember when the log4j/JNDI thing was a big fire, it seemed like the security commnity decided that it would be very difficult to patch all the applications that consume this feature without knowning or understanding how it worked - so the library interface should be updated to change the default.  Maybe this is only superficially similar.  I don't think they even did a major version bump, and I guess XML parsing is hardly as ubiquitous as logging.

But I do see a hint of some precedent in this project - https://github.com/lxml/lxml/commit/1a1031638330f508d3c61be7fb38fc41f8bd768a

Do you think it would be better if the default resolve_entities kwarg could change?

### DIFF
--- a/doc/FAQ.txt
+++ b/doc/FAQ.txt
@@ -1082,6 +1082,9 @@ question.
 How do I use lxml safely as a web-service endpoint?
 ---------------------------------------------------
 
+Good on you for finding this FAQ, but lxml make using XML in
+web-services safe by default.
+
 XML based web-service endpoints are generally subject to several
 types of attacks if they allow some kind of untrusted input.
 From the point of view of the underlying XML tool, the most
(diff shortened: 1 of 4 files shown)


clayg:
I also see now this is a duplicate of #130 (note the low bug number).  Software is so terrible.  We should give up.

scoder:
Closing as duplicate of #130.

PR #372, opened 2023-06-06
Add security policy

pnacht:
Related to https://bugs.launchpad.net/lxml/+bug/2023080.

This PR adds a security policy. The policy simply points users to use Launchpad and select the checkbox if they've found a security vulnerability.


scoder:
That's helpful. Thanks.

PR #375, opened 2023-06-28
stylistic changes to workflow file

rotu:
1. runner.os is enum-like. use `== 'macOS'` instead of `startsWith(runner.os, 'mac')`
2. instead of parsing `github.ref`, use github.ref_type
3. github action is weakly typed, so `== 'true'` is redundant


scoder:
Thanks

PR #378, opened 2023-07-04
Update error message to include envvar

rotu:
The build flags are all passable as environmnet variables (presumably since you can't pass flags with pip install). Update error message to make this more discoverable


scoder:
Thanks

PR #380, opened 2023-07-14
Skip test_isoschematron.test_schematron_invalid_schema_empty without the RNG file

hroncok:
The expected SchematronParseError only happens when validate_schema is true.


scoder:
Ah, right. Thanks. Clearly, the conditional code isn't actually exercised in the CI tests.

PR #382, opened 2023-10-18
Replace broken tutorial links with archive.org

isatsam:
These tutorials are no longer available at their original addresses:
- John Shipman's tutorial on Python XML processing with lxml
- Fredrik Lundh's tutorial for ElementTree
Replaced with most recent available Wayback Machine copies from 2010 and 2020 respectively


scoder:
Thanks

PR #383, opened 2023-10-22
Make Unicode recovery test work with upcoming libxml2 2.12

nwellnhof:
When encountering encoding errors, the upcoming version of libxml2 won't switch to ISO-8859-1 but replace illegal bytes with U+FFFD Replacement Character.


scoder:
Thanks. I'm happy about this change. It was a major source of weird errors, confusion and code quirks before.

PR #384, opened 2023-10-23
nh3 + new note into clean_html documentation

frenzymadness:
As described in https://bugs.launchpad.net/lxml/+bug/1958539/comments/8


scoder:
Thanks, looks good to me.

PR #385, opened 2023-10-30
Avoid usage of unittest.makeSuite, removed from Python 3.13

hroncok:
Patch generated by:

    sed -i 's/unittest.makeSuite/unittest.defaultTestLoader.loadTestsFromTestCase/g' $(grep -rl makeSuite)

No idea if this will work on Python 2, hence only opening as a draft.


scoder:
Thanks. I've already started working on the port, while testing Cython in 3.13. The 3.11 test failures are known.

I might consider finally dropping Py2.7 support in the upcoming lxml 5.0. Seems a good time to do it.

scoder:
THanks

PR #387, opened 2023-11-25
MNT: support for Python without PEP393 dropped by Cython

tacaswell:
This is the minimal change to get it to cythonize / compile again.

I am not sure if `python.PyUnicode_IS_READY(utext))` can ever be false (which would me some dead code could be ripped out), if the local variable should be renamed, and if dropping below py3.7 is on the table for lxml.

xref: https://github.com/cython/cython/pull/5801

### DIFF
--- a/src/lxml/parser.pxi
+++ b/src/lxml/parser.pxi
@@ -1029,8 +1029,7 @@ cdef class _BaseParser:
         cdef int buffer_len, c_kind
         cdef const_char* c_text
         cdef const_char* c_encoding = _PY_UNICODE_ENCODING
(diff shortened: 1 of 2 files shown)


scoder:
Thanks. It's probably still easier to just #define the PEP393_ENABLED macro to 1 if the Cython macro is not defined.

scoder:
This has been fixed in https://github.com/lxml/lxml/commit/3b545326cba14434649eee63bb828f80aa5276a4
To be released in lxml 5.0.

tacaswell:
Thank you!

PR #388, opened 2023-11-27
Build wheels with cibuildwheel

PrimozGodec:
In this PR, I am trying to build lxml with cibuildwheel.

It should fix the issue with missing arm wheels on Macos and simplify further management of wheels.

TODO:
- [ ] Tests
- [ ] Upload wheels to PyPI


.github/workflows/wheels.yml
```diff
@@ -70,187 +70,98 @@ jobs:
...
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.15.0
+        with:
+          only: ${{ matrix.only }}
+        env:
+          CIBW_SKIP: cp36-* cp37-* cp38-*
```
scoder, line 125: I'd be happy to keep those included. Adding back Py2.7 support would also be nice, but we could also keep the old workflow around for that.

scoder (commented):
Cool. That looks … ambitious. Let's see what it gives.

.github/workflows/wheels.yml
PrimozGodec: Thank you for your comment. I will add all versions/arch that you already have. This is just the first trial. :) 

I opened PR since I hoped I would see build runs, but they do not run without approval, so currently, I am checking the runs on the branch on my fork.

PrimozGodec:
Now, wheels are building well for all versions of Python 3.6+ and all platforms. I updated the description of the PR with TODO. At this point, I have two topics to discuss:
- lxml also provided Python 2.7 wheels. I believe it is not necessary anymore. Python 2 has not been supported since early 2020 (3 years from now) and is barely used. If one is brave enough to use Python 2 still, he/she can also build lxml from source (tar.gz from PyPI). This person must also build all other packages that they use since none of the major packages (numpy, pandas, sicpy, scikit-learn, ...) provides Python 2 wheels. Also, cibuildwheels, which Python Packaging Authority maintains, doesn't support Python 2.7 anymore. If lxml still wants to build Python 2.7, we must keep part of the old building process. But it is unnecessary since not having wheels doesn't mean the library doesn't support Python 2. It just means that the user must build it himself. What do you think?
- lxml wheels currently support a wide range of Python 3 versions and platforms, making it difficult to keep track of all. I suggest also stopping building all wheels for [Python versions that have already reached the end of life](https://devguide.python.org/versions/) as other major packages do (people who still want to use those old Python versions can build package themselves as they need to do for other packages). For example, currently, Python 3.8 is the latest supported. Some major packages went even further and stopped supporting 3.8 already (pandas, numpy), but I would suggest that lxml support all versions from 3.8 included.

scoder:
Here are the current download numbers per month:
```
Downloads by Python version:
  null:      91664.8 / day (  3.4%)
  3.13:         13.4 / day (  3.4%)
  3.12:       9466.7 / day (  3.8%)
  3.11:     107991.6 / day (  7.8%)
  3.10:     386333.3 / day ( 22.3%)
  3.9 :     320501.2 / day ( 34.3%)
  3.8 :     802057.9 / day ( 64.3%)
  3.7 :     765602.1 / day ( 92.9%)
  3.6 :     159096.6 / day ( 98.8%)
  3.5 :       7600.5 / day ( 99.1%)
  3.4 :         42.4 / day ( 99.1%)
  3.3 :         29.7 / day ( 99.1%)
  3.2 :          2.9 / day ( 99.1%)
  3.1 :          1.9 / day ( 99.1%)
  2.7 :      23384.3 / day (100.0%)
  2.6 :          2.9 / day (100.0%)
```
Dropping everything older than Py3.6 gives us >98% coverage. Assuming that most systems running Py<=3.5 will probably also not upgrade their lxml installation, I think it's safe to drop support for Py<3.6 in lxml 5.0 (i.e. soon).

scoder:
Thanks a lot, this is really sophisticated and helpful!

psahgal:
Hey, my team was using lxml in a Lambda function on Amazon Web Services and it looks like this new release broke our deployment pipeline. We were on Python 3.9. It looks like the image provided by Amazon that we're using to deploy no longer has a wheel to download. It attempts to build lxml from source and fails. Here's what I'm seeing in our build logs:

```
$ pip install lxml
Collecting lxml
  Downloading lxml-5.0.0.tar.gz (3.8 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3.8/3.8 MB 185.9 kB/s eta 0:00:00
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error

  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [3 lines of output]
      Building lxml version 5.0.0.
      Building without Cython.
      Error: Please make sure the libxml2 and libxslt development packages are installed.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.

[notice] A new release of pip is available: 23.0.1 -> 23.3.2
[notice] To update, run: pip install --upgrade pip
```

It's possible to reproduce the error if you've got Docker installed. Run `docker run -it --entrypoint bash public.ecr.aws/sam/build-python3.9:latest` and then `pip install lxml` and you'll see the error. This issue doesn't come up if you run `pip install lxml==4.9.4`.

I know this pull request removed some support for lxml on some configurations. Was the intention to remove support for the version of Linux in this build image? Here's the details on the Linux version for this image.
```
$ cat /etc/os-release
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
SUPPORT_END="2025-06-30"
VARIANT_ID="202312041833-2.0.1157.0"
```

In case anyone needs more details on what this image is used for, the [AWS docs](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-image-repositories.html) has a good explanation.

I'm upgrading us to Python 3.12, but it looks like anyone else using the AWS Serverless Application Model to run Lambda functions on 3.9 will no longer be able to use newer versions of lxml.

scoder:
> We were on Python 3.9. It looks like the image provided by Amazon that we're using to deploy no longer has a wheel to download. 

Does this problem persist? The wheel upload took a little longer this time, so maybe it was just temporary? Which wheel does your system download normally? Or which does it choose for 4.9.4?

Note that this PR did not contribute to 5.0.0. It's a preparation for 5.1. If the problem persists, please open a new ticket in the bug tracker.

psahgal:
It appears to be working now! Never mind then.

PR #394, opened 2024-01-07
Update xmlerror.pxd for libxml2 2.12

hmaarrfk:
They made this const I believe.

see https://github.com/GNOME/libxml2/blob/d2b55a7a029c47ebcfa6ede88331d38f90d3eb24/include/libxml/xmlerror.h#L864


scoder:
Thanks. It's interesting that I adapted all usages but forgot to change the declaration, apparently.

hmaarrfk:
Yes. I was also very confused about why I only had to change one file!!!!

PR #400, opened 2024-01-22
Typo fixes

nosnilmot:


scoder:
Oh… well. Type it wrong once, copy it everywhere. :)
Thanks for catching that.

PR #405, opened 2024-02-17
Replace factory functions with classes.

udifuchs:
Until now Element and ElementTree were factory functions. Element() returned an _Element class and
ElementTree() returned an _ElementTree class.

This PR turns them into classes. These classes should behave exactly the same as the factory function. Specifically, in both cases:
```
>>> element = Element("test")
>>> type(element)
<class 'lxml.etree._Element'>
>>> callable(Element)
True
```
The difference is that before this change we could not use isinstance:
```
>>> isinstance(element, etree.Element)
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```
While now:
```
>>> isinstance(element, etree.Element)
True
>>> issubclass(_Element, Element)
True
```
These checks work because _Element is registered as a virtual subclass of Element.

The motivation for this PR is to support type annotations.

Currently, type stubs for lxml need to annotate the factory functions like this:
```
def Element(...) -> _Element: ...
```
Furthermore, the _Element class members are all annotated. But _Element was supposed to be kept as an internal implementation.

With this PR, the Element class can be used as an interface class. This will allow us to create modified stubs for lxml where the _Element class is not mentioned at all.

As far as I can tell, this change should not break backward compatibility. Someone would have to go out of their way to test that Element is a function and not a class. As mentioned above, both are `callable`.

Still, I must admit that I am not convinced that this is the correct approach. Specifically, it seems a bit of a hack to have an Element class that ends up creating an _Element class.


udifuchs:
> @udifuchs any update on this?

I just updated my branch to be in sync with lxml:master. But I don't get to decide if it is going to be merged.

When I made this PR about a year ago, I had plans to publish a matching PR to types-lxml once this PR is accepted. Meanwhile, types-lxml went through a lot of changes. Updating my changes there is a bit more than just fixing conflicts. I will try to update it in the next few days and send it to types-lxml.

Generally, these two PR should be independent. But I don't think that it  make sense make Element a class in types-lxml, unless lxml makes this change first.

Also, types-lxml does some runtime checks of the types, but I still don't fully understand how these work and if it has any implications for my change.

udifuchs:
I created a PR for the types-lxml: https://github.com/abelcheung/types-lxml/pull/81

If the pull request to types-lxml is applied, users will be able to annotate using the public Element "class":
```
def print_tree(tree: e.ElementTree[e.Element]) -> None: ...
```
It would work even if this PR is not applied.

But without this PR, users will still need to use the private _Element for  type narrowing:
```
def print_element(element: e.Element) -> None: ...

def func(element: object) -> None:
    # isinstance(element, e.Element)   # This line will raise an Exception without this PR.
    isinstance(element, e._Element)  # The type of element is now narrowed to e._Element
    print_element(element)  # Function call is valid static typing
```

dpgraham4401:
Thanks @udifuchs, I wasn't aware of the [types-lxml](https://github.com/abelcheung/types-lxml) package, that's very helpful.

scoder (commented):
I'll accept this PR. Thanks.

Please move the test to `test_etree.py` (removing the unused import).

While I understand the reason for keeping the indentation in the original `ElementTree` function, please change it to use four spaces throughout.

scoder:
I ran the [etree benchmark](https://github.com/lxml/lxml/blob/f56aa1de81a644189d6c4bbb3495043cdacc0500/benchmark/bench_etree.py) over your branch on Python 3.12. The tree setup times grow by 5-25%:
```
Current master:
     T1: 0.0155 0.0170 0.0177 0.0152 0.0172 0.0174
     T2: 0.0164 0.0191 0.0193 0.0184 0.0208 0.0215
     T3: 0.0036 0.0045 0.0040 0.0146 0.0154 0.0154
     T4: 0.0001 0.0001 0.0001 0.0003 0.0003 0.0003

lxe: create_elements              (--TC T1   )    0.0286 msec/pass, best of (   0.0288    0.0286    0.0291)
lxe: create_elements              (--TC T2   )    0.5600 msec/pass, best of (   0.5641    0.5600    0.5634)
lxe: create_elements              (--TC T3   )    0.0031 msec/pass, best of (   0.0031    0.0033    0.0036)
lxe: create_elements              (--TC T4   )    0.0281 msec/pass, best of (   0.0281    0.0281    0.0281)
lxe: create_subelements           (--TC T1   )    0.0253 msec/pass, best of (   0.0260    0.0253    0.0255)
lxe: create_subelements           (--TC T2   )    0.4935 msec/pass, best of (   0.4983    0.4935    0.4942)
lxe: create_subelements           (--TC T3   )    0.0029 msec/pass, best of (   0.0029    0.0031    0.0031)
lxe: create_subelements           (--TC T4   )    0.0253 msec/pass, best of (   0.0257    0.0253    0.0260)

PR branch:
     T1: 0.0161 0.0183 0.0182 0.0158 0.0188 0.0186
     T2: 0.0174 0.0232 0.0217 0.0234 0.0253 0.0247
     T3: 0.0040 0.0047 0.0045 0.0172 0.0170 0.0165
     T4: 0.0001 0.0001 0.0001 0.0003 0.0003 0.0004

lxe: create_elements              (--TC T1   )    0.0298 msec/pass, best of (   0.0298    0.0298    0.0300)
lxe: create_elements              (--TC T2   )    0.5920 msec/pass, best of (   0.5949    0.5941    0.5920)
lxe: create_elements              (--TC T3   )    0.0036 msec/pass, best of (   0.0036    0.0036    0.0036)
lxe: create_elements              (--TC T4   )    0.0293 msec/pass, best of (   0.0296    0.0298    0.0293)
lxe: create_subelements           (--TC T1   )    0.0272 msec/pass, best of (   0.0272    0.0279    0.0272)
lxe: create_subelements           (--TC T2   )    0.5312 msec/pass, best of (   0.5312    0.5329    0.5319)
lxe: create_subelements           (--TC T3   )    0.0031 msec/pass, best of (   0.0031    0.0033    0.0033)
lxe: create_subelements           (--TC T4   )    0.0269 msec/pass, best of (   0.0269    0.0269    0.0269)
```

Given that the goal of this PR is pretty much code cosmetics, I find a performance loss in that order rather high for a library that aims for high-speed XML processing.

src/lxml/etree.pyx
```diff
@@ -62,6 +62,8 @@ from libc cimport limits, stdio, stdlib
 from libc cimport string as cstring_h   # not to be confused with stdlib 'string'
 from libc.string cimport const_char
 
+from abc import ABC
```
scoder, line 65:
```suggestion
cdef object ABC
from abc import ABC
```

src/lxml/etree.pyx
```diff
@@ -3216,30 +3233,34 @@ def SubElement(_Element _parent not None, _tag,
...
+                return result_container.result
+        else:
+            c_doc = _newXMLDoc()
+            doc = _documentFactory(c_doc, parser)
+
+        return _elementTreeFactory(doc, element)
 
-    return _elementTreeFactory(doc, element)
+# Register _ElementTree as a virtual subclass of ElementTree
+ElementTree.register(_ElementTree)
```
scoder, line 3265:
```suggestion
# Remove "ABC" helper from module dict again
del ABC
```

src/lxml/etree.pyx
```diff
@@ -62,6 +62,8 @@ from libc cimport limits, stdio, stdlib
 from libc cimport string as cstring_h   # not to be confused with stdlib 'string'
 from libc.string cimport const_char
 
+from abc import ABC
+
```
scoder, line 66:
```suggestion
```

src/lxml/etree.pyx
```diff
@@ -3092,18 +3094,33 @@ cdef xmlNode* _createEntity(xmlDoc* c_doc, const_xmlChar* name) noexcept:
 
 # module-level API for ElementTree
 
-def Element(_tag, attrib=None, nsmap=None, **_extra):
+
+class Element(ABC):
```
scoder, line 3098:
```suggestion
from abc import ABC

class Element(ABC):
```

scoder:
I was surprised myself that the numbers were that bad, given that we're only changing `Element()` and not `SubElement()`, which should largely dominate the tree building. So I triple checked and found that I had run the benchmarks in debug mode ~ sorry for that.

I did the timings again and it looks much better:
```
Current master:
     T1: 0.0160 0.0153 0.0149 0.0132 0.0149 0.0161
     T2: 0.0150 0.0164 0.0172 0.0169 0.0192 0.0192
     T3: 0.0034 0.0036 0.0034 0.0148 0.0150 0.0160
     T4: 0.0001 0.0001 0.0001 0.0005 0.0004 0.0003

lxe: create_elements              (--TC T1   )    0.0215 msec/pass, best of (   0.0224    0.0224    0.0215)
lxe: create_elements              (--TC T2   )    0.4103 msec/pass, best of (   0.4108    0.4103    0.4120)
lxe: create_elements              (--TC T3   )    0.0021 msec/pass, best of (   0.0024    0.0024    0.0021)
lxe: create_elements              (--TC T4   )    0.0205 msec/pass, best of (   0.0212    0.0207    0.0205)
lxe: create_subelements           (--TC T1   )    0.0215 msec/pass, best of (   0.0215    0.0310    0.0231)
lxe: create_subelements           (--TC T2   )    0.4170 msec/pass, best of (   0.4199    0.4170    0.4344)
lxe: create_subelements           (--TC T3   )    0.0024 msec/pass, best of (   0.0024    0.0029    0.0024)
lxe: create_subelements           (--TC T4   )    0.0215 msec/pass, best of (   0.0215    0.0215    0.0215)

PR branch:
     T1: 0.0137 0.0155 0.0159 0.0136 0.0155 0.0162
     T2: 0.0156 0.0174 0.0180 0.0180 0.0195 0.0204
     T3: 0.0034 0.0038 0.0037 0.0147 0.0149 0.0151
     T4: 0.0001 0.0001 0.0001 0.0003 0.0003 0.0003

lxe: create_elements              (--TC T1   )    0.0231 msec/pass, best of (   0.0236    0.0231    0.0234)
lxe: create_elements              (--TC T2   )    0.4590 msec/pass, best of (   0.4933    0.4590    0.4714)
lxe: create_elements              (--TC T3   )    0.0052 msec/pass, best of (   0.0052    0.0052    0.0055)
lxe: create_elements              (--TC T4   )    0.0236 msec/pass, best of (   0.0455    0.0238    0.0236)
lxe: create_subelements           (--TC T1   )    0.0229 msec/pass, best of (   0.0229    0.0231    0.0234)
lxe: create_subelements           (--TC T2   )    0.4358 msec/pass, best of (   0.4554    0.4358    0.4492)
lxe: create_subelements           (--TC T3   )    0.0041 msec/pass, best of (   0.0041    0.0052    0.0052)
lxe: create_subelements           (--TC T4   )    0.0219 msec/pass, best of (   0.0241    0.0241    0.0219)
```

So, the regression in tree building (the first block of timings) is always in the order of ~5%, but you can see the big difference for `Element()` in the `create_elements` micro-benchmark, which is much slower now (> 20%).

I think a 5% slow-down in realistic code is probably acceptable for this, but the regression is still rather unfortunate. There's certainly code out there that will see a real difference.

udifuchs:
The difference between master and this branch is that instead of calling the `Element` function, cpython is calling the `Element.__new__` class method. My guess is that the slowdown is from cpython having to inspect the `Element` class.

If cython classes could have a `__new__` method, it might have been able to accelerate this. But I don't have a deep understanding of cython. I guess that you are the expert...

scoder:
I tried the benchmark again, this time using Cython 3.1a1+ (current master branch). For the previous runs I used the (almost latest) 3.0.11 stable release 

The results look much better again:
```
Current master:
     T1: 0.0151 0.0155 0.0155 0.0137 0.0156 0.0157
     T2: 0.0155 0.0168 0.0171 0.0175 0.0203 0.0215
     T3: 0.0047 0.0039 0.0035 0.0162 0.0149 0.0153
     T4: 0.0001 0.0001 0.0001 0.0003 0.0003 0.0003

lxe: create_elements              (--TC T1   )    0.0217 msec/pass, best of (   0.0217    0.0219    0.0217)
lxe: create_elements              (--TC T2   )    0.4268 msec/pass, best of (   0.4375    0.4268    0.4303)
lxe: create_elements              (--TC T3   )    0.0029 msec/pass, best of (   0.0029    0.0029    0.0029)
lxe: create_elements              (--TC T4   )    0.0217 msec/pass, best of (   0.0417    0.0224    0.0217)
lxe: create_subelements           (--TC T1   )    0.0212 msec/pass, best of (   0.0215    0.0215    0.0212)
lxe: create_subelements           (--TC T2   )    0.4232 msec/pass, best of (   0.4239    0.4590    0.4232)
lxe: create_subelements           (--TC T3   )    0.0024 msec/pass, best of (   0.0024    0.0024    0.0026)
lxe: create_subelements           (--TC T4   )    0.0212 msec/pass, best of (   0.0215    0.0212    0.0212)

PR branch:
     T1: 0.0133 0.0151 0.0155 0.0132 0.0148 0.0151
     T2: 0.0148 0.0170 0.0172 0.0167 0.0191 0.0192
     T3: 0.0030 0.0039 0.0037 0.0143 0.0149 0.0144
     T4: 0.0001 0.0001 0.0001 0.0005 0.0003 0.0003

lxe: create_elements              (--TC T1   )    0.0362 msec/pass, best of (   0.0362    0.0424    0.0424)
lxe: create_elements              (--TC T2   )    0.4439 msec/pass, best of (   0.4439    0.4444    0.4442)
lxe: create_elements              (--TC T3   )    0.0029 msec/pass, best of (   0.0029    0.0029    0.0029)
lxe: create_elements              (--TC T4   )    0.0243 msec/pass, best of (   0.0248    0.0243    0.0248)
lxe: create_subelements           (--TC T1   )    0.0236 msec/pass, best of (   0.0236    0.0236    0.0241)
lxe: create_subelements           (--TC T2   )    0.4199 msec/pass, best of (   0.4265    0.4199    0.4344)
lxe: create_subelements           (--TC T3   )    0.0024 msec/pass, best of (   0.0024    0.0026    0.0024)
lxe: create_subelements           (--TC T4   )    0.0212 msec/pass, best of (   0.0219    0.0212    0.0219)
```
Here, the results are much more mixed and definitely in the same ball park. That's most likely due to the improved support for the Python vectorcall procotol in Cython 3.1, specifically for calls with keyword arguments and kwargs dicts.

I'll postpone the next lxml release (6.0) a little more and try to get out Cython 3.1b1 in the meantime, so that I can base lxml 6.0 at least on that.

scoder:
After running the benchmarks a couple of times, the micro benchmark shows a regression of about 5% in Python 3.14 and 8-10% in Python 3.12. I haven't tried older Python versions but they are more likely to perform worse than better.

I'd be happy if this was an easier choice. On the one hand, this is a considerable performance regression without any runtime benefits. On the other hand, it is only for `Element()`, which is unlikely to make a visible difference in real world code. Well written code should call `SubElement()` or use other ways to efficiently create elements directly in the target document tree. (By design, `Element()` must always create a new document.)

Given how large lxml's API is, I can understand the urge to get usage help and completion directly from IDEs.

Therefore, I accept this PR into lxml 6.0. Thanks @udifuchs.

PR #406, opened 2024-02-27
Use html.clean from external project and provide html_clean extra

frenzymadness:
As discussed in https://bugs.launchpad.net/lxml/+bug/1958539 `lxml.html.clean` is now ready as a separate project `lxml_html_clean`: 

* Github: https://github.com/fedora-python/lxml_html_clean
* PyPI: https://pypi.org/project/lxml-html-clean/
* Documentation: https://lxml-html-clean.readthedocs.io/en/latest/#

I thought about direct dependency and extra and chose extra, which means:

* Projects that use lxml without `lxml.html.clean` will continue to use it without any difference. Except they won't have potentially vulnerable code installed.
* Projects that need `lxml.html.clean` will need to switch their requirements from `lxml` to `lxml[html_clean]`. That's all.

We do have a list of projects we know about using the `lxml.html.clean` and we opened issues for them in August last year so I can just ping them there about this change.

I'd like to add a changelog entry but I don't know which release you want this to be included in.


frenzymadness:
Tests are failing, that's expected for now. I can fix them either by adding lxml_html_clean to the test dependencies or by removing the tests of the removed code. I think the latter is better so this project does not test code from another one.

frenzymadness:
New changes:

- test_autolink will be a part of the new project soon, see https://github.com/fedora-python/lxml_html_clean/pull/4
- Add setuptools to test environments with Python 3.12+
- Fixed one unnecessary import of html.clean in test_basic
- Add lxml_html_clean as a test dependency.

The last point is important. Because test_feedparser actually uses HTML cleaning inside, it might make sense to add lxml_html_clean as a test dependency and use that test as a smoke test for backward-compatible import of the moved module.

scoder:
Thanks, looks very good. I'll merge it for the next non-bugfix release.

Note that the new project no longer compiles the code with Cython, so it will be slower than the binary `clean` module that lxml provided. Can't say how important this is for users, so I'm just mentioning it as a difference. It's certainly some overhead to start distributing binary wheels for various Python versions. Maybe you can at least wait and see how well the Limited API support will work in the upcoming Cython 3.1, so that you can ship one binary wheel instead of a whole range. We're working on it.

frenzymadness:
Thank you for the confirmation. That's my plan exactly - wait for Cython 3.1 and try to cythonize the clean module into the stable ABI. If that won't work, I can configure CI to produce the whole matrix of wheels.

frenzymadness:
Do you have any plans for the next (non-bugfix) releases? To know whether we should expect this to be shipped this month/quarter/year…

scoder:
> Do you have any plans for the next (non-bugfix) releases? To know whether we should expect this to be shipped this month/quarter/year…

I'm currently working up a bug-fix release and wouldn't mind sending out a 5.2 soon afterwards that only removes the cleaner. Not sure how long this will take, but hopefully some time in April.

scoder:
Thank you for taking this over!

PR #407, opened 2024-02-27
Fix test_elementtree with Expat 2.6.0

hroncok:
Feeding the parser by too small chunks defers parsing to prevent CVE-2023-52425. Future versions of Expat may be more reactive.

Heavily inspired by https://github.com/python/cpython/commit/4a08e7b3431cd32a0daf22a33421cd3035343dc4

We cannot use a @fails_with_expat_2_6_0 decorator
because the test passes in ETreePullTestCase.

Co-authored-by: @serhiy-storchaka

### DIFF
--- a/src/lxml/tests/test_elementtree.py
+++ b/src/lxml/tests/test_elementtree.py
@@ -10,6 +10,7 @@ import copy
 import io
 import operator
 import os
+import pyexpat
 import re
 import sys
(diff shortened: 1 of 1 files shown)


scoder:
Thanks

kloczek:
Is it possible to release new version because this PR has been merged? 🤔

PR #409, opened 2024-03-07
set python_requires >= 3.7

graingert:
you'll need to yank 5.1.0 so that older pythons can install lxml again


graingert:
my mistake this is incorrect

PR #413, opened 2024-03-28
FIX: Allow multiple inlined image data links in html clean

beledouxdenis:
Add a lazy quantifier in the regex `_find_image_dataurls` to match as few characters as possible,
to make it stop at the first occurence of `;base64,`

e.g.
```py
>>> _find_image_dataurls = re.compile(r'data:image/(.+);base64,', re.I).findall
>>> _find_image_dataurls('<div style="background: url(data:image/jpeg;base64,foo); background-image: url(data:image/jpeg;base64,foo);"></div>')
['jpeg;base64,foo); background-image: url(data:image/jpeg']
```

```py
>>> _find_image_dataurls = re.compile(r'data:image/(.+?);base64,', re.I).findall
>>> _find_image_dataurls('<div style="background: url(data:image/jpeg;base64,foo); background-image: url(data:image/jpeg;base64,foo);"></div>')
['jpeg', 'jpeg']
```

This allows to have multiple image data links on the same line, which happens for instance in inline styles.

Without this change, `_has_javascript_scheme` returns `True` because the count of safe image urls is lower than the number of possible malicious scheme.
Then, the whole style is dropped as considered malicious.

### DIFF
--- a/src/lxml/html/clean.py
+++ b/src/lxml/html/clean.py
@@ -54,7 +54,7 @@ _looks_like_tag_content = re.compile(
 # All kinds of schemes besides just javascript: that can cause
 # execution:
 _find_image_dataurls = re.compile(
-    r'data:image/(.+);base64,', re.I).findall
+    r'data:image/(.+?);base64,', re.I).findall
 _possibly_malicious_schemes = re.compile(
     r'(javascript|jscript|livescript|vbscript|data|about|mocha):',
     re.I).findall
--- a/src/lxml/html/tests/test_clean.py
+++ b/src/lxml/html/tests/test_clean.py
@@ -256,6 +256,31 @@ class CleanerTest(unittest.TestCase):
                 cleaned,
                 "%s  ->  %s" % (url, cleaned))
 
+    def test_image_data_links_in_inline_style(self):
+        safe_attrs = set(lxml.html.defs.safe_attrs)
+        safe_attrs.add('style')
+
+        cleaner = Cleaner(
+            safe_attrs_only=True,
+            safe_attrs=safe_attrs)
+
+        data = b'123'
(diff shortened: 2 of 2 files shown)


scoder:
Thanks. Note that the cleaner is currently being extracted from lxml, so the best target for this PR is here:
https://github.com/fedora-python/lxml_html_clean

beledouxdenis:
I will open a pull request in that repository then. Do you have a policy for bug fixes in stable releases ?

scoder:
> Do you have a policy for bug fixes in stable releases ?

Not yet, I think. It's an independent project with its own release cycle now. Can probably just go in.

beledouxdenis:
> > Do you have a policy for bug fixes in stable releases ?
> 
> Not yet, I think. It's an independent project with its own release cycle now. Can probably just go in.

I didn't mean in that independent project, but here. Any chance to see this bug fix land in lxml 5.1.0 or even earlier version, given it's a bug fix and not an improvement ?

scoder:
> Any chance to see this bug fix land in lxml 5.1.0 or even earlier version, given it's a bug fix and not an improvement ?

Hmm. I'd actually prefer waving this away, given that I'm planning to remove l.h.clean in lxml 5.2, hopefully soon. No-one complained about this for a decade or so, so I'd like to avoid any potentially disruptive changes to existing release series at this point. You consider this a bug fix, but for others, it's an unexpected change in their output.

beledouxdenis:
> You consider this a bug fix, but for others, it's an unexpected change in their output.

Well, at Odoo we have exactly the same concern, so I understand your point.
We will monkey-patch this then !

scoder:
Closing since `lxml.html.clean` has now been removed from lxml.

beledouxdenis:
Superseded by fedora-python/lxml_html_clean#5.

PR #416, opened 2024-04-01
Build: fix Makefile to work on systems without `which(1)`

mgorny:
Replace the use of third-party `which(1)` executable with the shell `command -v` builtin.  This avoids unnecessary dependency on a third-party executable, in favor of a shell builtin that is guaranteed by POSIX, and therefore fixes building documentation on systems that no longer installs `which(1)` by default, such as Gentoo Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ ftest_inplace: inplace
 	$(PYTHON) test.py -f $(TESTFLAGS) $(TESTOPTS)
 
 apidoc: apidocclean inplace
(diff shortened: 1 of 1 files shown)


scoder:
Thanks. Probably ok, but I'm still triggering a fresh build to make sure everything builds.

scoder:
I'll add this to 5.2.1.

mgorny:
Thanks a lot!

PR #417, opened 2024-04-04
Skip test_feedparser_data if lxml_html_clean is not available

frenzymadness:
This is useful mostly for distributors shipping lxml without lxml_html_clean.

I did it the same way as it's done for cssselect in test_css.py.


scoder:
Thanks.

PR #418, opened 2024-04-04
There is no -manylinux- in cp36-manylinux_x86_64

hroncok:


scoder:
Ah, thanks. I hadn't seen this before fixing it myself, so it's already changed.

PR #421, opened 2024-05-05
Remove check for ']]>' in CDATA class, add test

gertjanklein:
CDATA containing CDEnd (']]>') is handled by libxml2 by splitting it into ']]' and '>'. It is no longer needed to disallow CDend. A test to ensure this is added.


scoder:
Thanks

PR #423, opened 2024-05-15
Fix `KeyError` in `Schematron` When Handling Missing or Invalid Namespace Prefixes

DaveLak:
Fixes: https://bugs.launchpad.net/lxml/+bug/2058177

Resolves an issue in the `lxml.isoschematron.Schematron` class where an unhandled KeyError was raised when `Schematron` is initialized with elements that have invalid or missing namespace prefixes. The KeyError occurred due to direct dictionary access without checking for the existence of the key.

The changes introduced here:

- Replace direct dictionary access with dict.get() to safely retrieve values from namespace maps, preventing KeyError on missing/invalid prefixes.
- Add unit tests to cover cases where schemas lack proper namespace prefixes or contain incorrect namespace prefixes.

### DIFF
--- a/src/lxml/isoschematron/__init__.py
+++ b/src/lxml/isoschematron/__init__.py
@@ -225,7 +225,7 @@ class Schematron(_etree._Validator):
         schematron = None
         if element.tag == _xml_schema_root:
             schematron = self._extract_xsd(element)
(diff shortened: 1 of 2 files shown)


scoder:
Looks good to me. Thanks.

PR #426, opened 2024-06-19
Allow smart strings to be disabled on CSSSelector

apasel422:


apasel422:
These selectors can only return elements, so this is actually unnecessary.

PR #437, opened 2024-10-07
add more elements to lxml.html.builder

npt:
Based on [the list on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element). I was missing \<section\>, \<figure\>, \<main\>, etc. while using it.


scoder:
Thanks

PR #440, opened 2024-10-25
tests: Adjust HTML tests for upcoming libxml2 changes

nwellnhof:
Starting with 2.14.0, the tokenizer in libxml2's HTML parser conforms to HTML5 which means that some documents will be parsed differently. The following changes affect the lxml test suite:

- <title> is an RCDATA element, meaning that it can only be closed with a "</title>" string. This is easy to fix by closing the element correctly. Unfortunately, libxml2's old behavior is used to document and test how the parser previously recovered from malformed HTML.

- Newlines are normalized now and U+000D CARRIAGE RETURN (CR) is stripped from the input.

- ASCII control chars are now allowed.

- Processing instructions are parsed as bogus comments.

Instead of trying to check for different libxml2 versions, I simply deleted parts of some tests.

### DIFF
--- a/doc/parsing.txt
+++ b/doc/parsing.txt
@@ -246,7 +246,7 @@ this feature.
 
 .. sourcecode:: pycon
 
-  >>> broken_html = "<html><head><title>test<body><h1>page title</h3>"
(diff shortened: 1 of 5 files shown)


scoder:
Thanks!

PR #442, opened 2024-10-30
remove flat_namespace link flag on darwin

paparodeo:
darwin uses a two-level namespace by default and linking with flat_namespace changes the runtime symbol lookup algorithm to resolve symbols at runtime with the first definition found rather than the dylib used at link time. flat_namespace causes issues when more than one library is loaded into the address space.

### DIFF
--- a/setupinfo.py
+++ b/setupinfo.py
@@ -352,13 +352,6 @@ def cflags(static_cflags):
             if not possible_cflag.startswith('-I'):
                 result.append(possible_cflag)
 
-    if sys.platform in ('darwin',):
-        for opt in result:
-            if 'flat_namespace' in opt:
-                break
-        else:
-            result.append('-flat_namespace')
-
     return result
 
 def define_macros():


scoder:
The flag used to be needed essentially forever. I can't say if it still is.

paparodeo:
we've had some issues with nix and other python libraries that link with flat_namespace and then end up resolving the wrong symbols. one of them being pycurl: https://github.com/pycurl/pycurl/pull/855

scoder:
The thing is, the flag was added [16 years ago](https://lxml.de/5.3/changes-5.3.0.html#beta3-2008-06-19) and since then, almost everyone seemed to be happy with it. So it's risky to just remove it now.

I recall that there were random crash reports at the time, possibly due to mixing (or some kind of interaction between) the statically linked libraries and the system-wide installed (usually much older) libraries. Although from [reading up on the option's documentation](https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html), it's not clear to me why that would happen *without* that flag.

That said, wheels are built with LTO these days, compilers are newer, systems are newer, library loaders are newer, system libraries are newer and different. It might be worth a new try. But trying something out with the macOS build usually means to make a release, wait for bug reports, ask users to debug it on their side (or to discuss with those who wanted the change if the world is better with or without the change), wait for someone to come along who has enough knowledge (or debug time) to explain the problem, keep motivating them until they find a solution, integrate the solution, make a new release, then repeat until the complaints stop. That's a bunch of annoying work.

I'll just wait for the CI wheel builds to be happy, accept your PR if so, and then see what happens.

paparodeo:
> I'll just wait for the CI wheel builds to be happy, accept your PR if so, and then see what happens.

sounds good. thanks!

PR #448, opened 2025-01-27
Allow fromstring with buffer objects

jsonn:
Database adapters like psycopg will use memoryview objects to represent binary strings. There is no need to force applications to copy the data to bytes manually as the logic already works fine for buffer objects. As such, just relax the type check.


scoder:
Thanks. I think generally checking for the buffer interface would be better than adding only memoryview.

jsonn:
AFAICT, memoryview is how the buffer interface is exposed as Python object at least in Python 3. In Py2, there was also "buffer", but upstream lxml no longer has to care about that.

PR #449, opened 2025-01-29
Add docstrings relating to functions involving parser context to parser.pxi

abepolk:
I wrote docstrings for a number of functions in `parser.pxi`. Since the docs [said](https://lxml.de/lxml-source-howto.html#lxml-etree) `parser.pxi` is "definitely not the right place to start reading lxml's source code," I thought it would be helpful to make it more accessible to newcomers.

In particular, I focused on functions that involved the parser context. This includes the context at the lxml/Cython level and the libxml2/C level. I wrote these based on notes I took as I was reading through `parser.pxi` for the first time and learning how lxml's parser wraps libxml2. I include a little bit of detail on libxml2 internals because it appears coupled to lxml and I don't anticipate any changes soon to the libxml2 code I mention.

Feedback is welcome.

### DIFF
--- a/src/lxml/parser.pxi
+++ b/src/lxml/parser.pxi
@@ -574,7 +574,7 @@ cdef class _ParserContext(_ResolverContext):
 
     cdef void _initParserContext(self, xmlparser.xmlParserCtxt* c_ctxt) noexcept:
         """
-        Connects the libxml2-level context to the lxml-level context.
+        Connects the libxml2-level context to the lxml-level parser context.
         """
         self._c_ctxt = c_ctxt
         c_ctxt._private = <void*>self
@@ -591,15 +591,6 @@ cdef class _ParserContext(_ResolverContext):
                 self._c_ctxt.nsNr = 0
 
     cdef int prepare(self, bint set_document_loader=True) except -1:
-        """
-        This method resets the parser context for a new parse. It also connects the
-        error log to the libxml-level ctxt by passing in _receiveParserError in as a
-        callback. Note: the HTML parser always has the libxml2 ctxt->sax not set to
-        null, so this always works. The ctxt->sax is set up in the libxml2 funciton
-        htmlInitParserCtxt in HTMLparser.c, and the lxml error log is always connected.
-        Then libxml2's SAX's serror is set to be the place errors are sent when schannel
-        is set to ctxt->sax->serror in xmlCtxtErrMemory in libxml2's parserInternals.c.
-        """
         cdef int result
         if config.ENABLE_THREADING and self._lock is not NULL:
             with nogil:
@@ -609,6 +600,12 @@ cdef class _ParserContext(_ResolverContext):
                 raise ParserError, "parser locking failed"
         self._error_log.clear()
         self._doc = None
+        # Connect the lxml error log with libxml2's error handling. In the case of parsing
+        # HTML, ctxt->sax is not set to null, so this always works. The libxml2 function
+        # that does this is htmlInitParserCtxt in HTMLparser.c. For HTML (and possibly XML
+        # too), libxml2's SAX's serror is set to be the place where errors are sent when
+        # schannel is set to ctxt->sax->serror in xmlCtxtErrMemory in libxml2's
+        # parserInternals.c.
         # Need a cast here because older libxml2 releases do not use 'const' in the functype.
         self._c_ctxt.sax.serror = <xmlerror.xmlStructuredErrorFunc> _receiveParserError
         self._orig_loader = _register_document_loader() if set_document_loader else NULL
@@ -702,15 +699,8 @@ cdef xmlDoc* _handleParseResult(_ParserContext context,
                                 xmlparser.xmlParserCtxt* c_ctxt,
                                 xmlDoc* result, filename,
                                 bint recover, bint free_doc) except NULL:
-    """
-    C-level xmlDoc* result is set to NULL if the parser cannot parse the document.
-    Keep in mind the libxml2-level wellFormed, which starts at 1, is set to 0 if the
-    parser may be able to parse the document, but there are errors that libxml2 has
-    to work through. lxml then uses this, along with some other logic, to determine
-    whether to set lxml-level well_formed to 0 and if that happens, set result to
-    NULL. Then, if result is NULL, raise a parse error. Some modifications to result
-    are made. Then the xmlDoc* result is returned.
-    """
+    # The C-level argument xmlDoc* result is passed in as NULL if the parser was not able
+    # to parse the document.
     cdef bint well_formed
     if result is not NULL:
         __GLOBAL_PARSER_CONTEXT.initDocDict(result)
@@ -722,6 +712,13 @@ cdef xmlDoc* _handleParseResult(_ParserContext context,
         c_ctxt.myDoc = NULL
 
     if result is not NULL:
+        # In libxml2 logic that has already run, the libxml2-level wellFormed (camel
+        # case), which starts at 1, is set to 0 if the parser MAY be able to parse the
+        # document despite errors that libxml2 has to work through. So, here in this
+        # if/elif/else block, lxml-level well_formed (snake case) follows the same
+        # pattern; lxml decides whether to set well_formed from 1 to 0, in part based on
+        # that libxml2 wellFormed. However, lxml chooses to customise its version of the
+        # variable using additional criteria in this if/elif/else block.
         if (context._validator is not None and
                 not context._validator.isvalid()):
             well_formed = 0  # actually not 'valid', but anyway ...
@@ -895,13 +892,6 @@ cdef class _BaseParser:
         self._events_to_collect = (event_types, tag)
 
     cdef _ParserContext _getParserContext(self):
-        """
-        This method creates a parser context if there isn't already one. If
-        creating a new parser context, this function configures it, creates a
-        libxml2-level parser context, connects the two, and makes some changes
-        to the libxml2 context. Then, it returns the parser context for the
-        parser.
-        """
         cdef xmlparser.xmlParserCtxt* pctxt
         if self._parser_context is None:
             self._parser_context = self._createContext(self.target, None)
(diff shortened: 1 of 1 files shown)


src/lxml/parser.pxi
```diff
@@ -687,6 +702,15 @@ cdef xmlDoc* _handleParseResult(_ParserContext context,
...
                                 bint recover, bint free_doc) except NULL:
+    """
+    C-level xmlDoc* result is set to NULL if the parser cannot parse the document.
+    Keep in mind the libxml2-level wellFormed, which starts at 1, is set to 0 if the
+    parser may be able to parse the document, but there are errors that libxml2 has
+    to work through. lxml then uses this, along with some other logic, to determine
+    whether to set lxml-level well_formed to 0 and if that happens, set result to
+    NULL. Then, if result is NULL, raise a parse error. Some modifications to result
+    are made. Then the xmlDoc* result is returned.
+    """
```
scoder, line 713:
```suggestion
    """
    Determine if we can consider the parsing successful.

    Handles different error cases, e.g. with 'recover=True' or validation.
    Frees a parsed document if it turns out not to be acceptable.
    Raises an appropriate parse error.
    """
```

src/lxml/parser.pxi
```diff
@@ -871,6 +895,13 @@ cdef class _BaseParser:
         self._events_to_collect = (event_types, tag)
 
     cdef _ParserContext _getParserContext(self):
+        """
+        This method creates a parser context if there isn't already one. If
+        creating a new parser context, this function configures it, creates a
+        libxml2-level parser context, connects the two, and makes some changes
+        to the libxml2 context. Then, it returns the parser context for the
+        parser.
+        """
```
scoder, line 904:
```suggestion
```

src/lxml/parser.pxi
```diff
@@ -901,6 +932,9 @@ cdef class _BaseParser:
         return self._push_parser_context
 
     cdef _ParserContext _createContext(self, target, events_to_collect):
+        """
+        This method configures the lxml-level parser.
+        """
```
scoder, line 937:
```suggestion
        """
        Create and configure an appropriate lxml-level parser context.
        """
```

src/lxml/parser.pxi
```diff
@@ -947,6 +981,10 @@ cdef class _BaseParser:
         return 0
 
     cdef xmlparser.xmlParserCtxt* _newParserCtxt(self) except NULL:
+        """
+        This method calls into libxml to configure the libxml2-level parser context,
+        among other things.
+        """
```
scoder, line 987:
```suggestion
        """
        Create a libxml2-level parser context.
        """
```

scoder (commented):
Thanks.

I believe that documentation/docstrings should not repeat code. (Some of your proposals really just state what the code does.) It should rather explain and motivate what's not obvious from the code, why it's done, why it's needed.

abepolk:
Thanks for the feedback! I'm happy to make the changes you suggested. But first, I want to ask whether there is a better way to make lxml's `parser.pxi` more accessible to newcomers. Would a diagram, or a gloss of some sort be helpful? Or a list of common variables and what they represent? I'd like to make it so that people can read `parser.pxi` and quickly learn how it works. Thanks again!

scoder:
> Would a diagram, or a gloss of some sort be helpful? Or a list of common variables and what they represent?

Since you digged into it, you're probably in the best position to explain what information you would have liked to have. Then we could think about how best to present it.

src/lxml/parser.pxi
abepolk: This is mostly correct. However, I think the sentence `Handles different error cases, e.g. with 'recover=True' or validation.` is too simple. For example the block

```
        elif recover or (c_ctxt.wellFormed and
                         c_ctxt.lastError.level < xmlerror.XML_ERR_ERROR):
            well_formed = 1
```
is not so much describing any particular kind of error as much as saying to carry over libxml2 errors unless in lxml level recovery mode or libxml lastError has a level higher than what seems to be an error level named `XML_ERR_ERROR`. And, it's important to understand the underlying roles of the variables `c_ctxt.wellFormed` and `well_formed` because they're different but made of the same two English words. (I mean English words as opposed to variable names or keywords.)

The reason my docstring was originally long is because the code itself is very complex. What do you think?

src/lxml/parser.pxi
abepolk: The reason I wrote `among other things` is that the lines
`                self._registerHtmlErrorHandler(c_ctxt)`
and
`        c_ctxt.sax.startDocument = _initSaxDocument`
are important and not related to creating the parser context. This is especially true with the line about SAX, which does many things.

src/lxml/parser.pxi
```diff
@@ -687,6 +702,15 @@ cdef xmlDoc* _handleParseResult(_ParserContext context,
                                 xmlparser.xmlParserCtxt* c_ctxt,
                                 xmlDoc* result, filename,
                                 bint recover, bint free_doc) except NULL:
+    """
+    C-level xmlDoc* result is set to NULL if the parser cannot parse the document.
```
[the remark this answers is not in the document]
abepolk: Thanks for the feedback! The problem is that `result = NULL` appears many times in `parser.pxi`. `result`'s being `NULL` is also checked in many different places using `result is NULL` and `result is not NULL` throughout `parser.pxi`. So, an in-place code comment would have to be duplicated >10 times. Do you agree?

abepolk:
I was thinking that instead of these docstrings, I could instead write an unofficial gloss or "guide" to reading the code in a separate article on Medium. My sense is that there is a lot of complexity in lxml and libxml2 and coupling between the two. An unofficial gloss or guide might be able to explain the complexity better. Any thoughts on that?

scoder:
> write an unofficial gloss or "guide" to reading the code in a separate article

That would be a separate article and many people won't expect such an article to exist (and be worth searching for) when starting to read the code. Comments in the code would certainly be more helpful.

I write "comments" because code concepts are better explained in comments than in docstrings. docstrings are for users of the library. Comments are for readers of the source code.

You can also add to this document:
https://lxml.de/lxml-source-howto.html
It's at least closer to the code than a "random article" on Medium.

src/lxml/parser.pxi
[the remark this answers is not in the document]
scoder: … or appear at the top of the function. But it's still a code comment, not a docstring. docstrings are written for users of the function. Code comments are written for readers of the code.

src/lxml/parser.pxi
scoder: So, if specifically the case in the code that you copied above is not clear, why not just add a comment to it?

src/lxml/parser.pxi
scoder: Well, there is a libxml2 function to create the context. If the creation was trivial, we wouldn't need this function here to create it.
```suggestion
        """
        Create and initialise a libxml2-level parser context.
        """
```
Better?

src/lxml/parser.pxi
abepolk: I'm happy to add a comment to it rather then a docstring. But where does this leave us on content? Should I put what I had in the original PR, with docstrings replaced by code comments? Or what I mention above in my previous PR comment (not to be confused with code comments)? Or does it matter? I just want to be sure you're okay with whatever I come up with.

src/lxml/parser.pxi
abepolk: I would advocate getting rid of the docstring entirely and replace it with comments in the right places. We could avoid the words "among other things" by just not putting a comment above the lines with "other things" (for example `c_ctxt.sax.startDocument = _initSaxDocument`).

abepolk:
That makes sense. I left a few comments on the specific suggested changes in the PR. Overall, the plan of making this PR about code comments rather than docstrings makes sense to me. Does this mean the content I originally wrote might be worth keeping? For example, the parts about how libxml2 internals affect lxml?

scoder:
> Does this mean the content I originally wrote might be worth keeping?

Since it seemed to help you, it probably is. If you split it into code comments in place, we'll see whether they are really helpful or redundant.

abepolk:
I converted a number of the docstrings to comments. I also reworded and removed a bunch. I tried to take your feedback into consideration. Let me know what you think! And, no problem if this PR doesn't work out—I will keep my notes for future times I come to read `parser.pxi`.

src/lxml/parser.pxi
```diff
@@ -687,6 +702,15 @@ cdef xmlDoc* _handleParseResult(_ParserContext context,
                                 xmlparser.xmlParserCtxt* c_ctxt,
```
SAIVEERENDER24, line 699: test:;quantum;peformanxe;operatabiity;goverence/crptography:\
@abepolk @scoder 
please use the coder to integrate and perform the operations that may solve all the issues that had been raised by the particular developer and that may solve all the issues intime
#458 
supported evedience that had been integrated and the that had been deployed by the developer in the time duration that may effect the time hours that causes and irruption of the code that may decode and performs all the actions and that creates
@masklinn 
@scoder 
pyrex is great but that may creates many bugs that Automatically creates one to the other managing all, the system preferences that may create all the coding values and that gives an existing value of code that agrees all the code sequences and that may cause code erruption between the line

src/lxml/parser.pxi
```diff
@@ -698,6 +712,13 @@ cdef xmlDoc* _handleParseResult(_ParserContext context,
         c_ctxt.myDoc = NULL
 
     if result is not NULL:
+        # In libxml2 logic that has already run, the libxml2-level wellFormed (camel
+        # case), which starts at 1, is set to 0 if the parser MAY be able to parse the
+        # document despite errors that libxml2 has to work through. So, here in this
+        # if/elif/else block, lxml-level well_formed (snake case) follows the same
+        # pattern; lxml decides whether to set well_formed from 1 to 0, in part based on
+        # that libxml2 wellFormed. However, lxml chooses to customise its version of the
+        # variable using additional criteria in this if/elif/else block.
```
scoder, line 721:
```suggestion
        # "wellFormed" in libxml2 is 0 if the parser found fatal errors. It still returns a
        # parse result document if 'recover=True'. Here, we determine if we can present
        # the document to the user or consider it incorrect or broken enough to raise an error.
```

scoder:
Thanks

abepolk:
Glad you merged this! Thanks!

PR #450, opened 2025-02-03
iterparse: ignore strip_cdata with HTML

ferdnyc:
Commit b79424c deprecated the `strip_cdata` argument to the HTML parser, causing all uses of `iterparse()` to trigger its `DeprecationWarning` (due to the default `True` value). Remove the `strip_cdata` argument from the HTML parser's arguments, and document it as ignored in `iterparse()` except for XML documents.

Tests are added to `test_htmlparser.py`, I've verified that they fail on the `master` branch without this change, but pass on the PR branch.

See also: https://bugs.launchpad.net/lxml/+bug/2067707


scoder:
Thanks. I'll release a 5.3.1 soon that should include this fix.

PR #452, opened 2025-02-04
CI: Remove ubuntu 20.04 job

ferdnyc:
GitHub is shutting down the ubuntu-20.04 runners. They'll all be gone by April 1, with "brownouts" (random failures) likely to start before then.

Ref: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#ubuntu-20-image-is-closing-down

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,9 +135,6 @@ jobs:
 
           # Legacy jobs
           # ===========
-          - os: ubuntu-20.04
-            python-version: "3.6"
-            env: { STATIC_DEPS: true }  # only static
           - os: ubuntu-22.04
             python-version: "3.7"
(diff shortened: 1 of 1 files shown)


ferdnyc:
(That ubuntu-22.04 build is failing due to commit c58e3de71d2a0c74a1c401698ab956d3ca787741 — it raised the libxml2 library version to 2.9.14, which works for ubuntu-24.04, but not for 22.04. Only 2.9.13 is available there.)

scoder:
Seems that removing the job is the best way forward. Thanks.

ferdnyc:
\*nod\* I subsequently discovered that the official deprecation period actually started Friday (or Saturday, or whenever the 1st was). It's time for sure.

PR #453, opened 2025-02-13
Additions to the documentation on how to build lxml

abepolk:
I wrote two minor additions to the documentation on how to build lxml. They are based on some things I had to figure out to build the library.

### DIFF
--- a/doc/build.txt
+++ b/doc/build.txt
@@ -135,10 +135,11 @@ files to the include path like::
 
 where the file is in ``/usr/include/libxml2/libxml/xmlversion.h``
 
-For static builds, if you get an error saying "recompile with -fPIC", do so with
-``env CFLAGS="-fPIC"``, such as::
+For static builds, if you get an error saying "recompile with -fPIC",
+do so by adding it to your `CFLAGS` environment variable:
+``env CFLAGS="$CFLAGS -fPIC"``, such as::
 
-  env CFLAGS="-fPIC" python3 setup.py build_ext -i --with-cython --static-deps
+  env CFLAGS="$CFLAGS -fPIC" python3 setup.py build_ext -i --with-cython --static-deps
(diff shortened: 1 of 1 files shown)


doc/build.txt
```diff
@@ -135,6 +135,11 @@ files to the include path like::
 
 where the file is in ``/usr/include/libxml2/libxml/xmlversion.h``
 
+For static builds, if you get an error saying "recompile with -fPIC", do so with
+``env CFLAGS="-fPIC"``, such as::
+
+  env CFLAGS="-fPIC" python3 setup.py build_ext -i --with-cython --static-deps
+
```
scoder, line 142: Right. I'd keep the original `CFLAGS` in there, though, just in case they are already set.
```suggestion
For static builds, if you get an error saying "recompile with -fPIC",
do so by adding it to your `CFLAGS` environment variable:
``env CFLAGS="$CFLAGS -fPIC"``, such as::

  env CFLAGS="$CFLAGS -fPIC" python3 setup.py build_ext -i --with-cython --static-deps

```

scoder (approved):

doc/build.txt
abepolk: This makes sense! Are you implying that the lxml user might have set `$CFLAGS` before they run the build command?

Also, did I hard wrap to the wrong number of chars?

scoder: Long lines are fine, there's no hard limit, more of a "feels better" limit.

And yes, users might have `CFLAGS` for one reason or another, willingly or by accident, and we should not ignore them to avoid unnecessary sources of confusion.

scoder:
Thanks

doc/build.txt
abepolk: That makes sense—thanks!

PR #454, opened 2025-02-14
Add a note about C compiler installation to error message

abepolk:
I wrote an update to a hard coded error I ran into trying to build lxml. The original error suggested that the problem might be with my libxml2 installation, but in fact I hadn't installed a C compiler.

### DIFF
--- a/setupinfo.py
+++ b/setupinfo.py
@@ -285,6 +285,7 @@ def print_libxml_error():
     print('Could not find function xmlCheckVersion in library libxml2. Is libxml2 installed?')
     if sys.platform in ('darwin',):
         print('Perhaps try: xcode-select --install')
+    print('Otherwise, is your C compiler installed and configured correctly?')
     print('*********************************************************************************')
 
 


setupinfo.py
```diff
@@ -285,6 +285,7 @@ def print_libxml_error():
     print('Could not find function xmlCheckVersion in library libxml2. Is libxml2 installed?')
     if sys.platform in ('darwin',):
         print('Perhaps try: xcode-select --install')
+    print('Otherwise, is your C compiler installed and configured correctly?')
```
scoder, line 288: Seems helpful, thanks. I'd move it up, though, before the macOS suggestion. (And I just noticed that the named C function is not the one we check for.)
```suggestion
    print("Could not find function xmlXPathInit in library libxml2. Is libxml2 installed?")
    print("Is your C compiler installed and configured correctly?")
    )
    if sys.platform in ('darwin',):
        print('Perhaps try: xcode-select --install')
```

setupinfo.py
```diff
@@ -282,7 +282,9 @@ def seems_to_have_libxml2():
 
 def print_libxml_error():
     print('*********************************************************************************')
-    print('Could not find function xmlCheckVersion in library libxml2. Is libxml2 installed?')
+    print("Could not find function xmlXPathInit in library libxml2. Is libxml2 installed?")
+    print("Is your C compiler installed and configured correctly?")
+    )
```
scoder, line 287:
```suggestion
```

scoder:
Thanks

setupinfo.py
abepolk: Makes sense! Glad to help.

PR #461, opened 2025-04-26
Enable subscripting generic types in annotations.

udifuchs:
Some classes like lxml.etree.ElementTree are generic types that can contain different kinds of elements. The type of these elements has to be specified in type annotations. For example:

element_tree: lxml.etree.ElementTree[lxml.etree.Element]

This requires adding a __class_getitem__ class method to these class, as specified in PEP 560. This commit adds these class methods.

The list of generic types in lxml was taken from:

https://github.com/abelcheung/types-lxml/wiki/Using-specialised-class-directly

This PR is a reincarnation of PR #401. The difference is that now there is no reference to private classes anymore.

### DIFF
--- a/src/lxml/builder.py
+++ b/src/lxml/builder.py
@@ -227,6 +227,10 @@ class ElementMaker:
     def __getattr__(self, tag):
         return partial(self, tag)
 
+    # Allow subscripting ElementMaker in type annotions (PEP 560)
+    def __class_getitem__(cls, item):
+        return f"{cls.__name__}[{item.__name__}]"
+
 
 # create factory object
 E = ElementMaker()
--- a/src/lxml/builder.py
+++ b/src/lxml/builder.py
@@ -229,6 +229,10 @@ class ElementMaker:
 
     # Allow subscripting ElementMaker in type annotions (PEP 560)
     def __class_getitem__(cls, item):
+        import sys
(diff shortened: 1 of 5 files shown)


scoder:
Hmm, is it still a good idea to return a string from `__class_getitem__`, now that we have [`types.GenericAlias`](https://docs.python.org/3/library/stdtypes.html#types-genericalias)? Admittedly, that's only available from Py3.9 onwards, but it's probably ok to return strings (or anything, really) in Py3.8 and use `GenericAlias` for everything since.

udifuchs:
I will look into the `GenericAlias` and update this PR.

scoder:
Rather than doing a version check on each use, in all `__class_getitem__` methods, I'd use something like
```python
cdef object _GenericAlias  # only in .pyx files, to store it outside of the module dict
try:
    from types import GenericAlias as _GenericAlias
except ImportError:
    # Python 3.8 - we only need this as return value from "__class_getitem__"
    def _GenericAlias(cls, item):
        return f"{cls.__name__}[{item.__name__}]"
```

udifuchs:
I made the change you requested.

Originally, I was trying to follow your "lazy import" code style. But eventually, this is only executed once per annotation when the code gets loaded, so I don't think it makes much difference for performance.

scoder:
Thanks!

PR #463, opened 2025-05-12
Improve html diff

scoder:
Supersedes https://github.com/lxml/lxml/pull/350


lonetwin:
@scoder yes indeed, I noticed the change! However, I didn't really pay close attention to the speed difference. Additionally, the test samples that I checked against are just a collection of small snippets extracted from various unit tests (from code at my work). ...a bit similar to those in the lxml codebase. I doubt the change would have been noticeable in the isolated runs I did. 

I can set something to benchmark the difference over multiple runs with and without the change and will report back.

PR #465, opened 2025-06-24
Add Windows arm64 wheel builds and tests

finnagin:
Fixes https://bugs.launchpad.net/lxml/+bug/2004481

This would add windows arm64 wheel builds to wheels.yml and tests to ci.yml.


finnagin:
FYI, it looks like some of the tests for macos and ubuntu in the CI workflow are [currently failing in master](https://github.com/lxml/lxml/actions/runs/15734026598) when `STATIC_DEPS` is `true` and I saw the same when running [the workflow on my fork,](https://github.com/finnagin/lxml/actions/runs/15858771698) though all of the windows tests (both the old x64 and added arm64) are passing thankfully.

PR #466, opened 2025-06-26
Make the XML names dict local to a parser

scoder:
This makes it easier to control for users than a dict that is local to a thread.


vesterbrookFF:
Hi!
I’m seeing some issues with the release of lxml 6.0.0 that’s blocking our CI/CD pipeline. We were using version 4.6.3, and after upgrading to 6.0.0, we’re still encountering failures. I’ve updated both my local machine and the repo to use the newer version, but the issue persists.

Could you advise on the next steps or any workarounds?

Thanks!

PR #467, opened 2025-07-04
Fix expected messages for XPath namespace errors

nwellnhof:
libxml2 2.15 adds additional detail to some XPath error messages, for example the actual name of undefined prefixes, functions or variables. Use the doctest.ELLIPSIS option to handle such messages.


scoder:
Thanks. Setting the general option seems fine for documentation text.

PR #469, opened 2025-07-23
Do not require wheel for building

hroncok:
- current version of setuptools (70.1+) does not need wheel at all
- older versions of setuptools would fetch wheel when building wheels (but not sdists)


scoder:
Thanks. Looks like the build failure is due to the CI change from Python 3.14b4 to rc1. I'll see that I get that sorted out, probably just outdated test assumptions.

PR #471, opened 2025-08-03
Feature/allow rv64 package

ffgan:
PyPI recently added support for RISC-V, so we might consider adding support for RISC-V64.

Below are the CI results:
[CI](https://github.com/ffgan/lxml/actions/runs/16694262589)
Like the current master build, it failed the Python 3.14 test (presumably unrelated to RISC-V64).

[Wheel build](https://github.com/ffgan/lxml/actions/runs/16699810497)
The results are generally well presented.

**Other Info**
Co-authored by: nijincheng@iscas.ac.cn;

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -161,7 +161,6 @@ jobs:
           CIBW_MANYLINUX_AARCH64_IMAGE: manylinux_2_28
           CIBW_MANYLINUX_PPC64LE_IMAGE: manylinux_2_28
           CIBW_MANYLINUX_S390X_IMAGE: manylinux_2_28
-          CIBW_MANYLINUX_RISCV64_IMAGE: ghcr.io/rust-cross/manylinux_2_31-cross:riscv64
           CIBW_MANYLINUX_PYPY_X86_64_IMAGE: manylinux_2_28
           CIBW_MANYLINUX_PYPY_AARCH64_IMAGE: manylinux_2_28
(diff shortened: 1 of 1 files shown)


scoder:
Thanks. Do you have an actual need for this or is this a "because we can" kind of PR? Building wheels on emulated CPUs takes a long time and lots of resources that I'd like to avoid wasting.

ffgan:
> Thanks. Do you have an actual need for this or is this a "because we can" kind of PR? Building wheels on emulated CPUs takes a long time and lots of resources that I'd like to avoid wasting.

I'm actually currently porting Ceph to RV64. One of the dependencies required for installing Ceph is lxml, and building a wheel from sdist is quite time-consuming, so I submitted this PR.
Of course, if you have any other questions, I'd be happy to help.

ffgan:
Hello  @scoder , I would like to respectfully inquire about the reasoning behind skipping `Python 3.9, 3.10, and 3.11` for `manylinux_riscv64` builds. I notice that these versions can also complete builds and tests successfully, so I'm curious if there are any specific considerations or technical reasons for this decision that I might not be aware of?

I have a need for riscv64 wheel for these three Python versions (3.9, 3.10, 3.11) when working with other packages like `python-xmlsec`. Would it be possible to consider adding support for these versions? 

I'd appreciate any insights you could share about this approach.

mattip:
Would it be an option to release a pure-python wheel for risc-v? Then there could be one wheel for all versions of python. BTW, this could also be done for PyPy.

PR #480, opened 2025-09-24
Avoid data buffer copying when writing to files

scoder:
Passes a thin `Py_buffer` wrapper object into the file-like `write()` method call instead of a 'bytes' copy.


scoder:
Manually merged in https://github.com/lxml/lxml/commit/819cdd9ab31ddad2e4a01d508b868cc185986137 and https://github.com/lxml/lxml/commit/9a1334251662e16785be6e40bb26a09c6f7aea97

PR #492, opened 2026-03-14
fix: htmldiff with proper error handling.

jucajuca:


src/lxml/html/diff.py
```diff
@@ -208,10 +208,13 @@ def htmldiff(old_html, new_html):
     new_html_tokens = tokenize(new_html)
     result = htmldiff_tokens(old_html_tokens, new_html_tokens)
     try:
-        result = ''.join(result).strip()
+        result = ''.join(result)
```
scoder: This doesn't seem to affect the error handling, but it does impact the result, which users might rely on. I think this part of the change should be reverted.

jucajuca:
Errors when the diff fails:

ERROR 2026-04-20 17:49:43,953 views 11 139708701178752 htmldiff failed joining result chunks: sequence item 917: expected str instance, type found

ERROR 2026-04-20 17:49:46,610 views 8 139708701178752 htmldiff failed joining result chunks: sequence item 921: expected str instance, type found

ERROR 2026-04-20 17:49:48,776 views 10 139708701178752 htmldiff failed joining result chunks: sequence item 829: expected str instance, type found

scoder:
Thanks

PR #499, opened 2026-04-22
Build: pin all actions

NyanKiyoshi:
Workflows are currently inconsistent, a lot of them are already pinned while some of them are not pinned yet. This looks accidental and thus this commit fixes the inconstency. The unpinned actions are also touching the release process which raises a concern that this could allow another compromised project to be able to poison the wheels generated by lxml (especially with less known/trusted parties like `softprops/action-gh-release`)

Note/credit: this PR was automated using https://github.com/suzuki-shunsuke/pinact (which allows to automatically pin all actions)


scoder:
Thanks

PR #501, opened 2026-04-29
Rewrite ElementPath as a Cython .pxi backed by a libxml2 walker

mahomaho:
Replaces the pure-Python ``lxml._elementpath`` module with two new
Cython files compiled into ``etree``:

* ``iterfind.pxi`` -- a self-contained C-level engine: a path
  tokenizer/compiler producing an array of step structs, a lazy
  depth-first cursor walker over libxml2 nodes (xmlNode*), and an
  ``_IterFindResult`` cdef class that yields one xmlNode* per call.
  Operates entirely in libxml2 land: no _Element wrapping inside the
  walker, no regex, no Python-level string juggling.

* ``_elementpath.pxi`` -- a thin Python-facing wrapper exposing
  ``find`` / ``findall`` / ``iterfind`` / ``findtext`` as cdef entry
  points that take an _Element, run the C engine, and call
  ``_elementFactory`` only on results actually returned to Python.

The compiler handles ``{uri}name`` Clark notation, ``prefix:name``
resolution against the namespaces dict (including default-namespace
binding via the None / '' key), and the path-syntax rules that the old
Python wrapper enforced separately (reject leading ``/``, treat
trailing ``/`` as implicit ``/*``). Predicates supported: ``[@attr]``,
``[@attr="value"]``, ``[@attr!="value"]``, ``[tag]``, ``[tag="value"]``,
``[tag!="value"]``, ``[N]``, ``[last()]``, ``[last()-N]``,
``[. = "value"]``, ``[. != "value"]`` -- the same set the Python
implementation supported.

The cursor walker is *lazy*: a path like ``item/name`` walks until the
first match and stops. ``find()`` short-circuits, ``findall()``
materialises only as it iterates. This was the single biggest fix --
an earlier eager-BFS prototype regressed ``find()`` 75x because it
materialised every descendant before returning the first.

The cursor struct is two pointers (scope, current); the three logical
states (fresh / in-progress / exhausted) are encoded by ``current``
alone using ``scope`` as the fresh sentinel. The steps and cursors
arrays share a single allocation. Predicates that compare strings go
through ``xmlGetNoNsProp`` / ``xmlNodeGetContent`` directly with no
Python copies.

## Performance

Measured against released lxml 6.1.0 (PYTHONPATH switching) on a 200x20
doc, 5000 iters per case, Python 3.14:

```
  case                                     old us  new us  speedup
  -----------------------------------------------------------------
  find("item")                              2.27    0.39   5.82x
  find(".//child")                          2.40    0.41   5.85x
  find("item/name")                         4.08    0.48   8.50x
  findtext("item/name")                     4.68    0.55   8.51x
  findall("item/name")                    296.27   71.30   4.16x
  findall("item/child")                   834.94  425.98   1.96x
  findall("item[@id]")                     74.79   30.40   2.46x
  findall("item[@type=\"odd\"]")           84.60   25.13   3.37x
  findall("item[1]")                     1553.77   86.12  18.04x
  findall("item[last()]")                1596.55    2.69 593.51x
  findall("item[name]")                   288.86   17.38  16.62x
  findall(".//child[@key=\"k5\"]")       1449.26  376.61   3.85x
  miss: findall("nope")                     3.60    1.43   2.52x

  findall(".//child")                     512.74  458.81   1.12x
  findall(".//*")                        1546.46  779.29   1.98x
```

The biggest wins (predicates) come from skipping the per-call Python
re-parse the old code did via the LRU cache hit; ``find()`` wins from
short-circuit + no factory calls on intermediate matches; descendants
are flatter because the bottleneck is libxml2 traversal itself.

## Validation

* test_elementpath: 28/28 OK on Python 3.11 and 3.14
* benchmark/verify_descendant.py: byte-identical findall() output
  vs released lxml across 8 descendant-style paths
* The xpath_tokenizer doctest in selftest.py (and a few tokenizer
  unit tests in test_elementpath.py) tested the *implementation* of
  the regex-based parser; they're skipped/removed since the new
  parser has no Python tokenizer to test.

## Files

* New: ``src/lxml/iterfind.pxi`` (~1100 lines), ``src/lxml/_elementpath.pxi``
* Removed: ``src/lxml/_elementpath.py``
* ``etree.pyx``: drop the ``from lxml import _elementpath`` import,
  ``include`` the two new pxi files, route the four call sites to the
  new cdef entry points
* ``setupinfo.py``: drop ``lxml._elementpath`` from COMPILED_MODULES
* ``selftest.py``, ``test_elementpath.py``: remove tests of the
  removed regex tokenizer
* ``benchmark/``: bench_compare, bench_elementpath, bench_find_scaling,
  bench_find_vs_iter, verify_descendant -- the benchmarks/cross-checks
  used to validate the rewrite

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


src/lxml/tests/selftest.py
```diff
@@ -958,7 +908,7 @@ def xmlwriter():
     >>> w.data("\n")
     >>> w.element("p", u"reserved characters: <&>")
     >>> w.data("\n")
-    >>> w.element("p", u"detta �r ocks� ett stycke")
+    >>> w.element("p", u"detta �r ocks� ett stycke")
     >>> w.data("\n")
```
scoder, line 909: This looks broken.

src/lxml/tests/test_elementpath.py
```diff
@@ -22,30 +22,16 @@ def normalize_crlf(tree):
 
 class EtreeElementPathTestCase(HelperTestCase):
     etree = etree
-    from lxml import _elementpath
 
     _empty_namespaces = None
 
+    @unittest.skip("internal _elementpath._cache API removed; replaced by C-level pxi")
     def test_cache(self):
```
scoder: This test can be deleted.

src/lxml/tests/test_elementpath.py
```diff
@@ -22,30 +22,16 @@ def normalize_crlf(tree):
...
-        self.assertTrue(el.findall('b/c'))
-        self.assertEqual(2, len(self._elementpath._cache))
+        pass
 
     def _assert_tokens(self, tokens, path, namespaces=None):
-        if namespaces is None:
-            namespaces = self._empty_namespaces
-        self.assertEqual(tokens, list(self._elementpath.xpath_tokenizer(path, namespaces)))
+        # The Python xpath_tokenizer is gone; this helper is unused by skipped tests.
+        self.skipTest("internal xpath_tokenizer API removed")
```
scoder: Method to be removed.

src/lxml/iterfind.pxi
```diff
@@ -0,0 +1,1114 @@
...
+        c_next = c_node.next
+        while c_next is not NULL:
+            if c_next.type == tree.XML_ELEMENT_NODE:
+                return c_next
+            c_next = c_next.next
+        c_node = c_node.parent
+
+    return NULL
+
+# ---- C-level path tokenizer / compiler ----
```
scoder, line 462: I don't think this necessarily needs to be done in C. A Python/Cython implementation should be fine and would probably be easier to maintain.

scoder (commented):
Thanks – that's a long chunk of code.

Did you compile the `_elementpath.py` module when running your benchmarks? It certainly makes a difference whether you're timing the pure Python module or the Cython compiled one.

A certain advantage of the current implementation is that it is shared with CPython's implementation (just faster), so changes on either side are easy to incorporate in the other. This is lost with a new implementation. I'm not generally opposed to a faster implementation (I always advocate `.find*()` over `.xpath()` because it's faster most of the time), but this loss is part of the decision.

To reduce the amount of added code and keeping it more maintainable, I suggest implementing the tokeniser and parser part in (compiled) Python, and only implementing the tree walker in C-ish code. Parsing shouldn't normally be performance critical and the result can be cached, as before.

lxml.etree already has a few tree walker implementations (like the `ElementDepthFirstIterator`) that can probably be reused and extended here. In any case, the implementation should reuse as much of the existing infrastructure as possible. The `_MultiTagMatcher` is definitely worth using, for example, since it pre-caches the tag name pointers.

Your new tree iterator does not currently hold a reference to Python `_Element` objects. It needs to keep the relevant parts of the current tree alive across iteration steps, though, and for that, it needs to instantiate at least one element and keep a reference to it. Look at the existing tree iterators to see how it's done.

I'm also suspicious about the lazy tree walking. We need to make sure that the tree walker doesn't get diverted (more than reasonable) when users remove parts of the current tree between iteration steps. It's certainly debatable where we draw the line, but it would be good to keep the bar low.

scoder:
> ```
>   findall("item[1]")                     1553.77   86.12  18.04x
>   findall("item[last()]")                1596.55    2.69 593.51x
>   findall("item[name]")                   288.86   17.38  16.62x
> ```

These are probably worth investigating separately. There might be easy wins in the current implementation (which CPython could possibly also benefit from).

src/lxml/tests/selftest.py
mahomaho: You are right. I will soon push a fix for it

src/lxml/tests/test_elementpath.py
mahomaho: Good, I will do so

src/lxml/tests/test_elementpath.py
mahomaho: done in next commit

src/lxml/iterfind.pxi
mahomaho: The reason why I started to look into this were due to some performance issues I have in a tool using lxml. This change alone saved me 25% of startup time, not from the big wins in the comparison but from the small gain cases, and that is the reason why I sent this in. You might be right that this does not need to be done c-style, my guess is that the big win here is to leave the old regex based solution and going into a solution that does not have to allocate a number of objects for each search.

src/lxml/tests/test_elementpath.py
```diff
@@ -339,15 +195,6 @@ def test_find_warning(self):
             etree.ElementTree(elem).iterfind, "/tag")
 
 
-class ElementTreeElementPathTestCase(EtreeElementPathTestCase):
```
scoder, line 355: This test class is for compatibility comparisons and uses `xml.etree.ElementTree` and its ElementPath implementation. It should stay in as a canary to notice functional divergence between both implementations.

src/lxml/iterfind.pxi
scoder: I think your observation that the repeated element instantiation is a costly part of the traversal is spot-on. It's probably the single most costly part of it. I've worked on that for the index predicates in https://github.com/lxml/lxml/pull/502

(Don't take that as rejecting your PR, it's just really low hanging fruit that can be done without replacing the whole implementation.)

mahomaho:
I am currently looking at the problem you pinpointed that a user may modify the lxml tree model after the iterator is created and before the iterator search has finished, leading to an access of freed memory. I hope that I will push a fix for this before the end of this week(the solution will be to save a reference to the last search result instead of the xml doc). I actually hope that this change will make the find even more efficient than my previous push

src/lxml/tests/test_elementpath.py
mahomaho: done

mahomaho:
I pushed a slightly modified implementation. Main difference is the handling of tree modifications while iter searcher is alive and not finished. Previously I used the doc reference to keep tree alive, now I use current search result.

The new implementation has improved speed with 10-20%

src/lxml/iterfind.pxi
```diff
@@ -0,0 +1,1170 @@
...
+
+        _if_ns_table_init(&nst)
+        try:
+            # Populate the table from the user's namespaces dict. The
+            # prefix/URI byte pointers point into each str's internal
+            # UTF-8 cache (PyUnicode_AsUTF8AndSize), so the str objects
+            # -- and therefore the dict that holds them -- must outlive
+            # the searcher chain. Iterators keep a copy on themselves;
+            # find/findall keep the dict alive via the function-local
+            # parameter.
```
scoder, line 924: A user provided data structure should be copied before relying on its content to not change. This problem is already solved in several other spots in lxml (tag matching, XPath extensions, XSLT parameters, …), but this one can probably be eliminated by not replicating the tag matching code.

src/lxml/iterfind.pxi
scoder: If I understand correctly, we lost the caching with your change, right? Meaning, it now parses the path each time it is evaluated? I'd rather go back to an LRU cache (mapping path strings to prepared evaluators) and keep the parser simple. For most applications, the parser performance really doesn't matter as long as each path is only parsed once (within reasonable cache size boundaries).

src/lxml/_elementpath.pxi
```diff
@@ -0,0 +1,138 @@
...
+# the namespaces dict. The iterator pins both via its own ``_path``
+# and ``_namespaces`` (shallow dict copy) fields; find/findall pin
+# them via the function-local parameters. ``_current_element`` on the
+# iterator additionally anchors the xmlDoc lifetime AND provides
+# ``_c_node`` as the leaf cursor for resume.
+
+
+@cython.final
+@cython.internal
+cdef class _ElementPathIterator(_IfSearcher):
```
scoder, line 44: Please try if you can base this class on `_ElementMatchIterator`. At least the `_MultiTagMatcher` should be usable. It pre-caches the tag name pointers to replace slow string comparisons with fast pointer comparisons. While it might not be perfectly designed for an incremental path search, it's probably still acceptable since its comparisons are really fast, possibly even fast enough to keep all tag names of the path in one matcher and adding a second level filter afterwards that checks if the match it found is what we expect at the current path level. Path expressions usually only contain very few different tag names with little risk of ambiguity.

scoder (commented):
I haven't looked through the details of the implementation, simply because it brings in way too much duplication of existing functionality. I do appreciate the goal of making the path evaluation fast, but any code duplication also means additional maintenance overhead in the longer term.

scoder:
BTW, I understand that this implementation is geared more towards performance than integration and maintainability. That is not a criticism of the implementation as such, but prevents me (the maintainer) from accepting it into mainline lxml in its current state.

In order to keep its current performance characteristics (and to keep them under your own control), one thing that you can do is create your own PyPI package that reimplements `ElementPath` on top of lxml for your own purposes. lxml has a C-API that allows getting at the underlying libxml2 data structure, which should be enough to get your implementation going. I consider the functionality sufficiently large, self-contained and independent to merit its own separate package.

That said, I'd be happy to integrate as much of the relevant improvements into mainline lxml as I can afford from a maintenance perspective, so further work on this PR is appreciated.

src/lxml/_elementpath.pxi
mahomaho: I hear what you say and I appreciate your feedback. I will first focus om performance. I can admit that I didn't expect any feedback at all when I pushed the PR. What I find interesting is the potential for performance improvements. I am actually working on another PR where I merge the libxml2 and python node/element object into one to improve performance. While working on that one I found area around find and since it was quite standalone I decided to try and create a standalone PR for that one.
My main target is an open source python package for Autosar configuration stored in xml. I am working on performance optimizations and next step would be to either go c/cython or do something around xml handling. I kind of like lxml and I have integrated many features from there so I decided to see what could be done and that is where I am right now.

I hope that I will get some time to investigate what you say and try out how it would impact the results. It is a simple thing to try out when you have help from AI, things that you used to think about trying out but never happended because there never were time enough can now be tried out in an afternoon.

mahomaho:
Hi again,

I have now had time to analyze your feedback, and I first must say that I appreciate your feedback and I totally understand your point about maintainability. Now to the details: I have looked at your suggestions about reuse and I do not see a straight path here and therefore I reach out to you for some feedback before I start to rewrite things.

I have analyzed the _ElementMatchIterator and it has major similarities with my iterator class. But there is a problem to resuse it if I am correct:
The class has a "walker" stateless function pointer to walk the model. It also has a filter test function to test if current node fulfills the request.

The problem is the stateless walker function. When doing a "xpath" search then the walker function depends on where you are in the search. Lets say that you have an xpath like "./a//b" then the walker shall first walk the shildren and search for a, then it shall walk all decendants and search for b. My solution for that is the linked list of searcher objects, one list element per level with different walker functions.

What I can do is to merge these classes. I would then delete my _ElementPathIterator and update the _ElementMatchIterator to something in the _ElementPathIterator direction. Second step would be to delete merge my searcher class with the _MultiTagMatcher. Finally I would update all points where these functions are used.

Is this what you are after? This will doulbe the size of the PR but I do not mind doing the job.

scoder:
The `_ElementMatchIterator` itself is intentionally limited. It's the base class for lxml's simple iterators that iterate children, siblings, ancestors. There is also the `ElementDepthFirstIterator` for deep tree traversal, which comes closer to what path traversal would need. I'd like a path based iterator to integrate with these classes, or at least with their design if something new is needed, with the intention to keep the overall complexity low (and knowing that path traversal is more complex than what is there).

> you have an xpath like "./a//b" then the walker shall first walk the shildren and search for a, then it shall walk all decendants and search for b. My solution for that is the linked list of searcher objects, one list element per level with different walker functions.

Yes, something like that is needed. It's also somewhat how the current Python generator based implementation works, it creates one generator per path construct/level and then chains them. The main issue there is that it processes Python Element objects. In order to implement a step-by-step path iterator, we need a series of "next node" traversal functions together with selector state. We already have most of the required iterators, but they always start from Python Element instances, which, again, we try to avoid here. The optimisation goal should be to push the selection into the libxml2 tree level and only instantiate tree nodes that match the whole path, not just intermediate parts of it.

Taking the "list of searcher objects" literally, what do you think about simply running down a Python list of Python searcher objects to process the path step by step? They can expose a `cdef` method that takes an `xmlNode*` and returns the next `xmlNode*`, which can then be passed into the method of the subsequent searcher object. A `NULL` return value makes it back up to the previous searcher. That keeps the traversal functionality and its selector configuration together (although the traversal could probably be extracted into a simple C function to make it reusable by itself – some already exist, some are part of iterator classes). And this seems a much simpler design than storing everything in low-level C structures. Certainly easier to extend.

scoder:
> only instantiate tree nodes that match the whole path, not just intermediate parts of it.

This is becoming even more interesting with the newly added locking in lxml 7.0, both to support Free-threading Python and to make thread concurrency generally more robust. The most visible side-effect is that Element object instantiation is somewhat slower, which hits the current ElementPath implementation quite hard, plus 10-30%, due to the sheer number of one-time visited elements. If the whole path traversal was done at the C node level, the whole tree could be locked once for reading and then traversed in one go, as is currently the case for XPath expressions (evaluated by libxml2 entirely).

scoder:
> Taking the "list of searcher objects" literally

Please see https://github.com/lxml/lxml/pull/511

I borrowed from your ideas but didn't reuse your code, I hope you don't mind. I really appreciate that you brought up the discussion.

scoder:
Closing since this is not going to get merged as it is.

PR #504, opened 2026-05-15
Fix typo: `type annotions` -> `type annotations` in PEP 560 comments

mokashang:
The comment `# Allow subscripting <Class> in type annotions (PEP 560)` was copy-pasted to four locations with the same typo (`annotions` instead of `annotations`). Fixed in `builder.py`, `parser.pxi` (twice) and `sax.py`.

Comment-only change, no behavioural impact.

### DIFF
--- a/src/lxml/builder.py
+++ b/src/lxml/builder.py
@@ -234,7 +234,7 @@ class ElementMaker:
     def __getattr__(self, tag):
         return partial(self, tag)
 
-    # Allow subscripting ElementMaker in type annotions (PEP 560)
+    # Allow subscripting ElementMaker in type annotations (PEP 560)
     def __class_getitem__(cls, item):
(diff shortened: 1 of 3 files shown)


scoder:
Thanks

cclauss:
I would avoid putting the misspelling in the commit title because many repos publish changelogs with the title of landed pull requests, which perpetuates the misspelling in the docs.  This means that typo discovery tools such as [`codespell`](https://pypi.org/project/codespell) will need to have that typo added to their `--ignore-words-list` which blinds those tools to future regressions.

scoder:
Oh well… thanks for noting this. :-/
OTOH, commit logs are usually generated and shouldn't be part of spell checks. Might not be easy to exclude in some cases, but I don't see value in including them.

PR #511, opened 2026-06-08
Reimplement ElementPath using libxml2 tree traversal

scoder:
Avoids Element object instantiation during the search and only instantiates matching Elements.


scoder:
Benchmark results are mixed. The master branch is substantially slower than before all the locking was applied. With this PR, some benchmarks are much faster, some even substantially faster than before the locking, others are still slower, some even slower than the current master.
https://github.com/lxml/lxml/actions/runs/27116556338?pr=511

PR #520, opened 2026-07-15
Avoid NameError in doctestcompare when parser fails

pctablet505:
Fixes https://bugs.launchpad.net/lxml/+bug/2143321

When temp_install() clones LXMLOutputChecker.check_output into the doctest runner's check function, the exception handler references etree.XMLSyntaxError. The old code set doctest.etree, which only works when the runner's check function globals happen to be `doctest.__dict__`. That is true for the stock OutputChecker, but not when the active runner uses a checker whose check_output is defined in another module (for example pytest's doctest plugin, or any OutputChecker subclass): there the clone runs against that other module's globals and raises NameError: name 'etree' is not defined.

Injecting etree directly into the check function's actual `__globals__` dict covers every invocation path.

Adds a regression test that drives the doctest through a checker defined outside the doctest module (so the failing path is actually exercised) and asserts the parse failure is reported as a normal doctest failure rather than raising NameError.

### DIFF
--- a/src/lxml/doctestcompare.py
+++ b/src/lxml/doctestcompare.py
@@ -404,8 +404,13 @@ def temp_install(html=False, del_module=None):
     check_func = frame.f_locals['check'].__func__
     checker_check_func = checker.check_output.__func__
     # Because we can't patch up func_globals, this is the only global
-    # in check_output that we care about:
-    doctest.etree = etree
+    # in check_output that we care about.  We inject it directly into the
+    # function's actual globals dict, which is not necessarily
+    # doctest.__dict__: when the active runner uses a checker whose
+    # check_output is defined in another module (e.g. pytest's doctest
+    # plugin or any OutputChecker subclass), those globals belong to that
+    # module instead.
+    check_func.__globals__['etree'] = etree
     _RestoreChecker(dt_self, old_checker, checker,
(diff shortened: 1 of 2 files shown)


src/lxml/doctestcompare.py
```diff
@@ -404,8 +404,13 @@ def temp_install(html=False, del_module=None):
...
     # Because we can't patch up func_globals, this is the only global
-    # in check_output that we care about:
-    doctest.etree = etree
+    # in check_output that we care about.  We inject it directly into the
+    # function's actual globals dict, which is not necessarily
+    # doctest.__dict__: when the active runner uses a checker whose
+    # check_output is defined in another module (e.g. pytest's doctest
+    # plugin or any OutputChecker subclass), those globals belong to that
+    # module instead.
+    check_func.__globals__['etree'] = etree
```
scoder, line 413: Adding `etree` to the `doctest` module seems bad enough for a hack but controllable. Adding it to an arbitrary user module might have unexpected side effects. I'd rather avoid that if there is another way to solve the problem at hand.

scoder:
Thanks for writing this up. I borrowed the test in https://github.com/lxml/lxml/commit/041746e11ef0acdc0a19ea9be77c803909260e46 but used a different fix.

PR #521, opened 2026-07-19
Fix bare except: use specific exception types in Python scripts

koteshyelamati:
Replace bare `except:` clauses with specific exception types in pure Python files:

- `doc/rest2html.py`: `except Exception:` (guards `locale.setlocale()` call)
- `download_artefacts.py`: `except Exception:` (guards `shutil.copyfileobj()` during download)
- `buildlibxml.py`: `except Exception:` (guards CPU count via `multiprocessing`)

Bare `except:` catches `BaseException` including `SystemExit` and `KeyboardInterrupt`, which can suppress intended program termination.

### DIFF
--- a/download_artefacts.py
+++ b/download_artefacts.py
@@ -70,7 +70,7 @@ def download1(wheel_metadata, dest_dir):
         try:
             with open(temp_file_path, "wb") as f:
                 shutil.copyfileobj(w, f)
-        except:
+        except Exception:
             if temp_file_path.exists():
                 temp_file_path.unlink()
             raise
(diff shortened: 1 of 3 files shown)


download_artefacts.py
```diff
@@ -70,7 +70,7 @@ def download1(wheel_metadata, dest_dir):
         try:
             with open(temp_file_path, "wb") as f:
                 shutil.copyfileobj(w, f)
-        except:
+        except Exception:
```
scoder: This one is intended.
```suggestion
        except:
```

koteshyelamati:
Thank you for the review and merge, @scoder!

